PR #15, opened 2022-12-17
Add support for strings, bytes, submessages and repeated fields in generator.

vesim987:


Arwalk:
Ok, but stage 2 compiler still doesn't work properly with the code, which has nothing to do with your PR in the end.

Thanks again!

PR #19, opened 2023-06-21
feat: ManagedStrings

### LINKED ISSUE #18: discussion: Managed strings
As of today, the library is leaking memory for user-allocated strings.

Tests are not failing because static/const strings are not accounted as leaks.

This could be fixed by recursively deallocating strings in the `.deinit` funciton but two problems may arise:
1. It should be extremely clear for the user that the ownership of string memory is now passed onto the proto message. Plus, the allocator for the string MUST be exactly the same as the one used to deinitialize the message.
2. Deallocating const strings may result in runtime exceptions

Proposal:

Replace the plain `[]const u8` by `ManagedString`. Keep the `?ManagedString` for optional semantics. `ManagedString` is an union with `OwnedString` and `ConstString`, the former is automatically deallocated, the later won't.

```zig

const ManagedStringKeys = enum {
  OwnedString,
  ConstString
};

const ManagedString = union(ManagedStringKeys) {
  // memory that needs to be allocated
  OwnedString: []const u8,
  // memory that is static/const
  ConstString: []const u8,
};

```

menduz:
closes #18


tests/tests_varints.zig
```diff
@@ -35,12 +35,3 @@ test "Varints - encode/decode equivalence" {
     defer decoded.deinit();
     try testing.expectEqualDeep(demo, decoded);
 }
-
-// Extracted from the documentation site
-// https://protobuf.dev/programming-guides/encoding/
```
menduz, line 40: This was moved to another file

Arwalk:
I finally could look a bit into this, sorry for the delay. I'm glad you took the initiative to merge it without me. Thanks again for your work. Everything's fine.

menduz:
I've been testing it with real use cases for several days without any single problem, I'm confident to launch a 1.0 version soon

PR #21, opened 2023-07-30
mark as "pub const" public imports

menduz:
If you are compiling the protocol buffers as a zig package, it may use only one entry point.

This change leverages `import public "somefile.proto"` to expose `pub const somefile = @include("somefile.pb.zig);` in the root file of the package. Effectively enabling access to all the user-defined sub-packages


Arwalk (approved):
Thank you for your work again.

PR #23, opened 2023-08-23
add `.dupe(allocator)` to every message and default values

menduz:
It recursively copies the message using a new allocator. Useful when working with arenas and multiple threads to prevent memory corruption.

Also introduces the `.Empty` string type and uses it for the default values of non-nullable strings when calling `.init(allocator)` or `.{}`


menduz:
@Arwalk this is ready to be reviewed

src/protobuf.zig
```diff
@@ -445,6 +463,79 @@ pub fn pb_init(comptime T: type, allocator: Allocator) T {
...
+    var result: T = undefined;
+
+    inline for (@typeInfo(T).Struct.fields) |field| {
+        @field(result, field.name) = try dupe_field(original, field.name, @field(T._desc_table, field.name).ftype, allocator);
+    }
+
+    return result;
+}
+
+/// Internal deinit function for a specific field
```
Arwalk, line 478:
```suggestion
/// Internal dupe function for a specific field
```

Arwalk (approved):
Ok, only a comment correction

PR #40, opened 2024-03-12
Generate services interfaces

### LINKED ISSUE #39: Can we also generate services along with outputting types?
Taking this proto file as an example:

https://github.com/lightningnetwork/lnd/blob/master/lnrpc/invoicesrpc/invoices.proto#L29

The rpc services functions are not getting generated. Only the types are. Is it possible to also add this?

Arwalk:
closes #39

### DIFF
--- a/bootstrapped-generator/main.zig
+++ b/bootstrapped-generator/main.zig
@@ -165,6 +165,7 @@ const GenerationContext = struct {
 
         try self.generateEnums(list, fqn, file, file.enum_type);
         try self.generateMessages(list, fqn, file, file.message_type);
+        try self.generateServices(list, fqn, file, file.service);
     }
 
     fn generateEnums(ctx: *Self, list: *std.ArrayList(string), fqn: FullName, file: descriptor.FileDescriptorProto, enums: std.ArrayList(descriptor.EnumDescriptorProto)) !void {
@@ -196,8 +197,8 @@ const GenerationContext = struct {
             return name;
     }
 
-    fn fieldTypeFqn(ctx: *Self, parentFqn: FullName, file: descriptor.FileDescriptorProto, field: descriptor.FieldDescriptorProto) !string {
-        if (field.type_name) |typeName| {
+    fn fieldTypeFqn(ctx: *Self, parentFqn: FullName, file: descriptor.FileDescriptorProto, field_name: ?pb.ManagedString) !string {
+        if (field_name) |typeName| {
             const fullTypeName = FullName{ .buf = typeName.getSlice()[1..] };
 
             if (fullTypeName.parent()) |parent| {
@@ -235,9 +236,7 @@ const GenerationContext = struct {
                 parent = parent.?.parent();
             }
 
-            std.debug.print("Unknown type: {s} from {s} in {?s}\n", .{ fullTypeName.buf, parentFqn.buf, file.package.?.getSlice() });
-
-            return try ctx.escapeFqn(field.type_name.?.getSlice());
+            return try ctx.escapeFqn(typeName.getSlice());
         }
         @panic("field has no type");
     }
@@ -327,7 +326,7 @@ const GenerationContext = struct {
             .TYPE_DOUBLE => "f64",
             .TYPE_FLOAT => "f32",
             .TYPE_STRING, .TYPE_BYTES => "ManagedString",
-            .TYPE_ENUM, .TYPE_MESSAGE => try ctx.fieldTypeFqn(fqn, file, field),
+            .TYPE_ENUM, .TYPE_MESSAGE => try ctx.fieldTypeFqn(fqn, file, field.type_name),
             else => {
                 std.debug.print("Unrecognized type {}\n", .{t});
                 @panic("Unrecognized type");
@@ -566,6 +565,7 @@ const GenerationContext = struct {
 
             try ctx.generateEnums(list, messageFqn, file, m.enum_type);
             try ctx.generateMessages(list, messageFqn, file, m.nested_type);
+            // no services in nested types apparently
 
             try list.append(try std.fmt.allocPrint(allocator,
                 \\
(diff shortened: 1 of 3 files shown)


Arwalk:
I've tried adding some cool comptime checks on the `init` method based on reflection of functions and all, but the truth is that they never get hit by the compiler. If a method doesn't exist, or misses a parameter, or doesn't have a right type somewhere, the compiler detects it before any of my checks.

So this should be enough to generate services interfaces.

Arwalk:
my brain is smooth, i forgot about error handling on the interfaces.

menduz:
I think we should get first a library that does RPC and then create the implementation on top of it. The PR may include a good approach, but I'd hesitate on committing to it without a real-world use case. 

Besides that, there are three concise things we should discuss:

1. More often than not, implementations of methods return `Future|Promise|AsyncHandles`. We may need to converge towards some managed structure that holds a value, a status and a callback. `Future(T) = struct { resolved: bool, value: ?T, callback: fn(error, T), fn deinit() }` 
2. Both streaming requests and responses are not yet implemented but are critical to grpc
3. We should think about implementing remote, transport and runtime errors

Arwalk:
> I think we should get first a library that does RPC and then create the implementation on top of it. The PR may include a good approach, but I'd hesitate on committing to it without a real-world use case.
> 
> Besides that, there are three concise things we should discuss:
> 
> 1. More often than not, implementations of methods return `Future|Promise|AsyncHandles`. We may need to converge towards some managed structure that holds a value, a status and a callback. `Future(T) = struct { resolved: bool, value: ?T, callback: fn(error, T), fn deinit() }`
> 2. Both streaming requests and responses are not yet implemented but are critical to grpc
> 3. We should think about implementing remote, transport and runtime errors

Rpc services defined in protobuf are not necessarily gRPC, and i do not plan on implementing gRPC in this repository. If i ever tackle the subject, it would be on another repository, build on top of this one. This PR's goal is to have at least some naive interface out of defined services, and end users are free to branch it to whatever protocol they prefer.

For async, coroutines are sadly still not back on the menu in the compiler, so i'd prefer to wait a stable release on this before trying to half-bake something that we'd have to maintain inside.

Error handling is discussed in the original thread, i plan on giving end users [this](https://github.com/Arwalk/zig-protobuf/commit/d162f1ffb4bacc1fa1072d57b74eee48fdebb81b), but implementations will obviously have to handle their own errors somehow.

I kinda agree with what you said, but it assumes all protobuf services uses gRPC and that's not the case according to the spec itself.

menduz:
> Rpc services defined in protobuf are not necessarily gRPC

I completely agree, and personally spent several months working on a protobuf-based RPC mechanism that is not gRPC either https://github.com/decentraland/rpc. 

My argument on this matter is that we should provide some mechanism to enable wrapping responses and stream request/responses to enable async capabilities. I cannot think about one single case of RPC that is not asynchronous (outside of microcontrollers with interrupts), for everything else keep plain old structs.

In pseudocode:

```diff
-pub const TestService = struct {
+fn TestService(comptime AsyncHandle: fn (anytype) type) type {
+ return struct {
    const Interface = struct {
-        Foo: *const fn (x: *anyopaque, p: FooRequest) FooResponse,
+        Foo: *const fn (x: *anyopaque, p: FooRequest) AsyncHandle(FooResponse),
    };

    _ptr: *anyopaque,
    _vtab: *const Interface,

    pub fn init(obj: anytype) TestService {
        const Ptr = @TypeOf(obj);
        const PtrInfo = @typeInfo(Ptr);
        std.debug.assert(PtrInfo == .Pointer); // Must be a pointer
        std.debug.assert(PtrInfo.Pointer.size == .One); // Must be a single-item pointer
        std.debug.assert(@typeInfo(PtrInfo.Pointer.child) == .Struct); // Must point to a struct
        const impl = struct {
-            pub fn Foo(ptr: *anyopaque, param: FooRequest) FooResponse {
+            pub fn Foo(ptr: *anyopaque, param: FooRequest) AsyncHandle(FooResponse) {
                const self: Ptr = @ptrCast(@alignCast(ptr));
                return self.Foo(param);
            }
        };
        return .{ ._ptr = obj, ._vtab = &.{
            .Foo = impl.Foo,
            .Bar = impl.Bar,
        } };
    }

-    pub fn Foo(self: TestService, param: FooRequest) FooResponse {
+    pub fn Foo(self: TestService, param: FooRequest) AsyncHandle(FooResponse) {
        return self._vtab.Foo(self._ptr, param);
    }
};
```

---

Another option is to specify that in the code generator, for project-idiomatic code generation

```zig
// build.zig
    .services = .{
      .async_wrapper_name = "AsyncWrapper",
      .async_wrapper_import = "some-lib",
     // generates something like `const AsyncWrapper = @import("some-lib");`
    },
});
```

---

A third option, and the one that reduces all uncertainty, is to expose services metadata as comptime const structs, like we do for fields.

Those can be traversed and used to generate code and to perform validations in a per-project basis.

```zig
// myService.pb.zig

...

pub const MyService = struct {
  const methods = .{
    const Foo = .{
      .request_type = FooRequest,
      .request_is_streaming = false,

      .response_type = FooResponse,
      .response_is_streaming = false,
    };
  };
};

```

Arwalk:
Yeah, i see you're point. I'm gonna look a bit more into it and how it's done in google's C++ interfaces for services. 
But i'm really bothered by the fact that async is currently not available in zig. I can't find a Future/Promise/AsyncResult in the std right now.

That said, i found https://github.com/Jack-Ji/zig-async which could be an inspiration (i don't agree with some of the approaches on it and i would really like to avoid relying on another library).

In the meantime, I still think a simple and naive synchronous interface could be useful somehow. I'll think a bit more about it.

menduz:
>I can't find a Future/Promise/AsyncResult in the std right now.

https://github.com/Jack-Ji/zig-async/blob/main/src/task.zig#L7 is perfect! Exactly as I was thinking about it. 


I'd say that we can pass in `Future` as comptime parameter to generate the service type.

```zig
// generated code
fn MyService(comptime ResponseWrapper: type) type {
    return struct {
      // ptr
      // vtable

      fn Foo(req: Request) -> ResponseWrapper(Response) {...}
    };
}

// user code for async response types using ZigFuture
const MyServiceAsync = pb.MyService(Future);

// user code for synchronous responses
const MyServiceSync = pb.MyService(Identity);
fn Identity(comptime T: type) type {
    return T;
}
```

> I still think a simple and naive synchronous interface could be useful somehow

I'd be all in on this approach if we can find those useful cases. Did you think about any?

Arwalk:
Very late answer, have been pretty busy lately

> I'd be all in on this approach if we can find those useful cases. Did you think about any?

As you said, there are the use case of simple devices with no real use for asynchronous patterns. Also i think it could be nice for server-side implementations to have it?

But it's kind of annoying to generate 2 different interfaces for the same thing, so i guess it makes more sens to work on a Future(T) interface, and propose a blocking version that should be easy to use for test and less evolved hardware/systems.

I guess i still have to think a bit more about it.

niteshbalusu11:
@Arwalk any luck with this PR?

Arwalk:
To avoid any confusion, and as i don't plan on working on this until async is back, i'm closing it.

PR #45, opened 2024-04-20
Update for zig 0.12.0

### LINKED ISSUE #44: Update to zig 0.12.0
Zig 0.12.0 has been released. Time to upgrade.

malcolmstill:
Closes #44 

- The only thing in zig 0.12.0 that looks like it requires a fix is the change to `@fieldParentPtr` that [now only takes 2 parameters](https://ziglang.org/download/0.12.0/release-notes.html#fieldParentPtr)
- `zig fmt` seems to have made a few changes to `build.zig` also
- I've updated the version in the Dockerfile to just 0.12.0, but I'm not sure if the intention is to use some `0.13.0-dev` version instead. I did leave the version as `master` in the GitHub action. Can change any of these as desired.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -3,6 +3,6 @@ RUN git config --global core.autocrlf input
 
 ADD . .
 
-RUN bash ./install-zig.sh 0.12.0-dev.3283+b8920bceb
+RUN bash ./install-zig.sh 0.12.0
 
 ENV AGREE=true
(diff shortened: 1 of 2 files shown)


Arwalk:
@malcolmstill perfect, it's fine to fix the version in there. Dockerfile here is supposed to include the "associated" version of the compiler to the code.

Leaving master in the github action is perfect: it runs every 15 days so i get a notification if something breaks, so i can make a new version.

Thank you for your contribution.

PR #46, opened 2024-06-14
Update build.zig for Zig 0.13.0

hans-tvs:
`cwd_relative` is relative to the cwd of the build runner process, which causes problems (missing files, as my build step looks in the wrong location for zig-protobuf's source code) when I try to import zig-protobuf as a module or add a proto-generating build-step using the exact instructions in the README.

This PR just replaces that with `b.path`, which makes those dependencies relative to your build.zig instead. I'm not intimately familiar with all your build steps, so it might be worth spot-checking that's semantically correct, but that change is sufficient for me to be able to import and use this in another Zig project in the manner described in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,10 +74,8 @@ pub fn build(b: *std.Build) !void {
     const gen_proto = b.step("gen-proto", "generates zig files from protocol buffer definitions");
 
     const protoc_step = protobuf.RunProtocStep.create(b, protobuf_dep.builder, target, .{
-        .destination_directory = .{
-            // out directory for the generated zig files
-            .path = "src/proto",
-        },
+        // out directory for the generated zig files
+        .destination_directory = b.path("src/proto"),
         .source_files = &.{
             "protocol/all.proto",
         },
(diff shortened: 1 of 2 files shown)


HendrikJanssen:
Interesting, I just now stumbled on that exact same issue!

I also think it would be nice to correct the `zig-cache` directory to the relocated `.zig-cache` dir in https://github.com/Arwalk/zig-protobuf/blob/9dd4eef382693296d7d7f3f8e35a76635aad4bdb/build.zig#L284

Also, the README.md should probably include the correct way to setup the gen-protoc step:

```zig
        .destination_directory = b.path("src/proto"),
        .source_files = &.{
            "protocol/all.proto",
        },
        .include_directories = &.{},
    });
```

hans-tvs:
Nice catch. I was wondering where that `zig-cache` was coming from, and I forgotten I'd made that same `destination_directory` change in my calling code.

Arwalk:
Thanks a lot, guys. I realise I haven't checked imports in an external project, so I'm not surprised this error appeared.

I'll check this ASAP.
I am very happy to know people use this library, and your comments and commits are appreciated.

Arwalk (approved):
Successfully imported the updated library on a toy/demo project, could run gen-proto, build and run successfully.

PR #47, opened 2024-06-15
fix: make paths relative to the owner for RunProtocStep

Natoandro:
`gen-proto` step fails when run from a separate project because the build system try to find the files in the current directory.

This PR addresses that issue by making paths relative to the source root.


Arwalk:
duplicate of #46 , but your interest and PR is extremely appreciated. Thank you for your work and time,

PR #49, opened 2024-06-30
JSON (de)serialization

### LINKED ISSUE #32: Json serialization/parsing
The official python library for protobuf from google has a [json format](https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html), and it is found in the [conformance tests proto definition](https://github.com/protocolbuffers/protobuf/blob/181bcee76d2639c30048746013ae5b64aad0e97a/conformance/conformance.proto#L37) so this could tie with #27 

Considering the zig library has a json parser, it shouldn't be too hard to do.

librolibro:
Closes #32

This one works without extra boilerplate code (some tests included) and I kinda like the solution, but still ...

To get things done I reimplemented a part of source code (see **innerParse** function in **std/json/static.zig**, the part with the .Struct case body). Since few internal non-public was needed i just copied it (eww).

Zig is on its early stages and standard library often changes - and that's why I partly don't like this PR because there might be lots or refactors (for example, standard token parser might change)

Feel free to judge it)


Arwalk:
Thanks again for everything. I'll do my best to check it out ASAP.

Arwalk:
Just saying that i haven't forgotten this, just trying to find some time to do it.

Arwalk:
Started reviewing it. You're still not forgotten :)

src/protobuf.zig
```diff
@@ -75,6 +76,19 @@ pub const ManagedString = union(ManagedStringTag) {
             else => {},
         }
     }
+
+    pub fn jsonParse(
```
Arwalk, line 80: Important: Add comment to show that it's used by std.json for parsing

src/protobuf.zig
```diff
@@ -75,6 +76,19 @@ pub const ManagedString = union(ManagedStringTag) {
...
+        allocator: Allocator,
+        source: anytype,
+        options: json.ParseOptions,
+    ) !ManagedString {
+        const string = try json.innerParse([]const u8, allocator, source, options);
+        return ManagedString.copy(string, allocator);
+    }
+
+    pub fn jsonStringify(self: *const ManagedString, jws: anytype) !void {
```
Arwalk, line 89: Important: Add comment to show that it's used by std.json for encoding

src/protobuf.zig
```diff
@@ -963,6 +977,91 @@ pub fn pb_decode(comptime T: type, input: []const u8, allocator: Allocator) !T {
...
+fn parseStructField(
+    comptime T: type,
+    result: *T,
+    comptime fieldInfo: StructField,
+    allocator: Allocator,
+    source: anytype,
+    options: json.ParseOptions,
+) !void {
+    // Is this a good way to check if it's ArrayList or not? Took from here:
+    // https://ziggit.dev/t/how-to-json-de-ser-struct-with-arraylist-field/3510/5
```
Arwalk, line 1020: Suggestion: Instead of trying to deduce it from the name of the type, `fieldInfo` could be of type `FieldDescriptor`. Then you'll just have to check if `ftype` is of type `.List` or `.PackedList`.

src/protobuf.zig
```diff
@@ -980,6 +1079,138 @@ pub fn MessageMixins(comptime Self: type) type {
...
+            return pb_json_decode(Self, input, options, allocator);
+        }
+        pub fn json_encode(
+            self: Self,
+            options: json.StringifyOptions,
+            allocator: Allocator,
+        ) ![]const u8 {
+            return pb_json_encode(self, options, allocator);
+        }
+        pub fn jsonParse(
```
Arwalk, line 1096: Important: Add comment to show that it's used by std.json for parsing

src/protobuf.zig
```diff
@@ -980,6 +1079,138 @@ pub fn MessageMixins(comptime Self: type) type {
...
+                        try source.skipValue();
+                    } else {
+                        return error.UnknownField;
+                    }
+                }
+            }
+            try fillDefaultStructValues(Self, &result, &fields_seen);
+            return result;
+        }
+        pub fn jsonStringify(self: *const Self, jws: anytype) !void {
```
Arwalk, line 1181: Important: Add comment to show that it's used by std.json for encoding

Arwalk (changes requested):
This is a very good start for something great. There are a few comments already placed in the review here, but i'll add this:

- How is OneOf case handled? Could you add a test (and code to make it work) ?
- I'm not a huge fan of using std.json "leaky" code without it being explicit to the end user. I think we need to use non-leaky methods and return a `Parsed(T)`, and at worst propose leaky json_decode on a separate interface (that would require ArenaAllocator instead of just Allocator).

Overall this is very good and my requests do not change the fact that i'm very thankful for your time and investment in this.

src/protobuf.zig
Arwalk: also need to handle oneOf here.

librolibro: That was a really good advice! Made it.

librolibro:
Made some commits.

All unions are handled by **std.json** internally, see [here](https://github.com/ziglang/zig/blob/1fc42ed3e7ca0b74b54aaa827276d995d6c7c6cd/lib/std/json/static.zig#L262) and [here](https://github.com/ziglang/zig/blob/1fc42ed3e7ca0b74b54aaa827276d995d6c7c6cd/lib/std/json/stringify.zig#L475) - it just worked out of the box, made some tests to prove it.

About using "leaky" versions of std.json (de)serialization functions - the only reason I chose them is because I didn't get it to work with non-leaky alternatives) I just remember than some tests blamed about memory leaks (std.testing.allocator is able to catch leaks) and some code didn't even compile, don't remember why tbh. I can try to explore it again with a fresh look.

And thanks for feedback, even the late one)

Arwalk:
T his is great. I'll have some fun with it and might even add some suggestions before approving it, but overall i'm positive this will be merged soon. Thanks again for your work.

librolibro:
LUL i just switched to a non-leaky function for JSON decoding and everything works :)

There are no memory leaks or compile errors that I've encountered before.

Arwalk:
@librolibro I added a few commits directly, i'll let you look at them and tell me if they're ok for you. According to the [json spec for protobuf](https://protobuf.dev/programming-guides/proto3/#json) `bytes` types are supposed to be base64 encoded.

We did not have a specific information for bytes, we had a shortcut to handle it just like strings. These last commits do not change the zig-facing interface of using ManagedStrings for bytes fields, but add the information needed in the `_desc_table` so that json encoding/decoding know when to do the the base64 soup.

I'll look tomorrow into the Map type, because i'm pretty sure we are not correct with the spec too (we handle it as a list of entries).

librolibro:
Everything looks clean at the first glance, good job at that! (**@embedFile** usage was nice, didn't know about it :))

Another thing that comes to my mind is the ability to **not add** fields into JSON string if their value is equal to default - e.g. Python's implementation of **google.protobuf.json_format.MessageToJson** comes with *including_default_value_fields* argument which if False by default. While specs say that

> A proto3 JSON implementation **MAY** provide the following options

and it's not a strict requirement it may be worth implementing this and since the message field iteration was already made in this code and not in **std.json** depths it shouldn't be that hard to do.

Arwalk:
I took the liberty to rebase after #51 was merged, there was a small conflict.

Arwalk:
We need to set ourselves some clear goals for this PR.

## Required TODO

- [ ] In json parsing/writing, Map fields should be written/read as json objects. Making it an actual map on the zig side will be another subject (#52)
- [ ] `Use proto field name instead of lowerCamelCase name`: according to the spec, we should convert field names to lowerCamelCase when writing, and accept lowerCamelCase and actual name when parsing. 

@librolibro if you start any of these subjects, tell me so we avoid doing it twice.

## Worth analyzing in separate issues (that will be created after the merge of this PR)

- [ ] Check support for mapping of types defined by google (Timestamp, Duration...)
- [ ] Pass options to the parser
  - [ ] to ignore unknown fields
  - [ ] to always emits fields without presence 
- [ ] Pass options to the writer
  - [ ] to support writing actual name and not the camelCase name
  - [ ] to support emitting enum values instead of names

librolibro:
I would try to implement field renaming to camelCase. The problem is json.ParseOptions is the only argument passes to innerParse internal function and I can't pass any additional options to my jsonParse methods. I can make field renaming as default since specs require it but without the alternative (which would be no case conversion at all) atm. If I understand correctly deserializer should handle both cases anyway (what if message contains two fields like **test_field** and **testField**? Is this a collision?)

I would also make default fields ignoring because it's logic will be in the same part of code.

P.S. About float handling - [this std.json test suite](https://github.com/ziglang/zig/blob/master/lib/std/json/JSONTestSuite_test.zig) covers a lot of edge cases and also not-a-numbers and infinities. I also saw usage of **std.fmt.parse_float** in std.json source code

Arwalk:
> I would try to implement field renaming to camelCase. The problem is json.ParseOptions is the only argument passes to innerParse internal function and I can't pass any additional options to my jsonParse methods. I can make field renaming as default since specs require it but without the alternative (which would be no case conversion at all) atm. If I understand correctly deserializer should handle both cases anyway (what if message contains two fields like **test_field** and **testField**? Is this a collision?)
> 
> I would also make default fields ignoring because it's logic will be in the same part of code.
> 
> P.S. About float handling - [this std.json test suite](https://github.com/ziglang/zig/blob/master/lib/std/json/JSONTestSuite_test.zig) covers a lot of edge cases and also not-a-numbers and infinities. I also saw usage of **std.fmt.parse_float** in std.json source code

For now, focus on accepting both camelCase and exact field names when parsing, and writing in camelCase. We will deal with optional behaviors later, but these are not optional.

Thanks for your investment ! Do not force yourself to do it, remember that you have no obligations here.

librolibro:
I just realized that your code (and also parts of Zig standard library) follows Zig style guide and mine is not :) Your base64 encoding/decoding snippets (despite following preferred style guide) looks foreign being surrounded my by code - I'll refactor that tomorrow, I'm too lazy atm.

Also i would write more tests: deserializing camelCase, snake_case and mixed strings, handling all kinds of .Optional's etc. I like how you put JSON string into separate file and then extracted it with **embedFile** call. This project structure comes to my mind:

```
zig-protobuf/
  generated/
  json_data/
    test1/
      struct_definition.zig (import generated message instance from it)
      camel_case.json
      snake_case.json
      mixed_case.json
```

The more I look into the code the less I like it tbh because in order to just change union's tag name I need to "copy" (and slightly modify) a part of code from **std.json** and there's lots of undocumented methods which might change (maybe even before 0.14.0 version bump) - but for now it's working which might be more important.

librolibro:
I understand why the last workflow failed (my new FieldNameCollision error is out of the expected set) but I can't reproduce it on my 0.13.0 version of the compiler. I know little about the error system in Zig - if I understand correctly now I need to pass error union to jsonParse function.

Or maybe it's better to remove this error "throwing" - I still don't how this situations should be handled:

```proto
syntax "proto3";

message BadMessageForSure {
    float field_with_collision = 1;
    float fieldWithCollision = 2;
}
```

It's PITA when it comes to JSON (de)serialization. Is it valid by protobuf after all? Maybe it's better to throw panic in such cases?

librolibro:
I made NaN/Infinity (de)serialization support but in order to make it i need to make 4 changes (encoding/decoding for single/repeated fields) - since you made .Bytes support only for single fields it means that **"repeated bytes"** fields are not supported ATM. I'll look into it.

Also the thing to check is the ability to parse integers/floats of any kinds not only from numeric JSON data types but also from strings (e.g. **1** and **"1"** should be interpreted equal when parsing **int32** field), spec says:

> Either numbers or strings are accepted. Exponent notation is also accepted.

So the more tests the better for sure (maybe it will work out of the box, who knows).

james-callahan:
With the current PR HEAD I get:
```
/home/user/.cache/zig/p/1220403b861068d46215af6e338e5acd5eefcbabec7f4b2518042c3ec0bc5dae6646/src/protobuf.zig:1472:54: error: no member named 'bytes' in '[32]u8'
            const innerAllocator = jws.nesting_stack.bytes.allocator;
                                                     ^~~~~
referenced by:
    write__anon_7032: /usr/lib/zig/std/json/stringify.zig:506:51
    stringify__anon_6045: /usr/lib/zig/std/json/stringify.zig:52:17
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
```

looking at the zig json/stringify.zig source, it looks as though `nesting_stack` can be different things depending on compile mode:
```

        nesting_stack: switch (safety_checks) {
            .checked_to_arbitrary_depth => BitStack,
            .checked_to_fixed_depth => |fixed_buffer_size| [(fixed_buffer_size + 7) >> 3]u8,
            .assumed_correct => void,
        },
        ```



It seems as though the allocator is just used for the camel case functionality; but that only needs to work at compile time.
So you should be able to swap to a comptime implementation:
```zig

fn to_camel_case(comptime not_camel_cased_string: []const u8) []const u8 {
    comptime var underscore_count: usize = 0;
    inline for (not_camel_cased_string) |byte| {
        if (byte == '_') underscore_count += 1;
    }
    if (underscore_count == 0 or underscore_count == not_camel_cased_string.len) {
        // You better not touch fields like "_____________" (or should we?)
        return not_camel_cased_string;
    }

    var capitalize_next_letter = false;
    var camel_cased_string: [not_camel_cased_string.len - underscore_count]u8 = undefined;

    var i: usize = 0;
    for (not_camel_cased_string) |char| {
        if (char == '_') {
            capitalize_next_letter = i > 0;
        } else if (capitalize_next_letter) {
            camel_cased_string[i] = std.ascii.toUpper(char);
            capitalize_next_letter = false;
            i += 1;
        } else {
            camel_cased_string[i] = char;
            i += 1;
        }
    }

    if (std.ascii.isUpper(camel_cased_string[0])) {
        camel_cased_string[0] = std.ascii.toLower(camel_cased_string[0]);
    }

    return &camel_cased_string;
}
```

From there it is a quick cleanup, aside from the base64 encoding, which needs to drop down to writing to the json stream directly rather than allocating a temporary.

src/protobuf.zig
```diff
@@ -963,6 +987,347 @@ pub fn pb_decode(comptime T: type, input: []const u8, allocator: Allocator) !T {
...
+                try json.innerParse(
+                    [](@typeInfo(fieldInfo.type.Slice).Pointer.child),
+                    allocator,
+                    source,
+                    options,
+                ),
+            );
+        },
+        .Bytes => bytes: {
+            // "bytes" -> ManagerString
```
james-callahan, line 1055:
```suggestion
            // "bytes" -> ManagedString
```

src/protobuf.zig
```diff
@@ -963,6 +987,347 @@ pub fn pb_decode(comptime T: type, input: []const u8, allocator: Allocator) !T {
...
+    } else if (std.math.isPositiveInf(value)) {
+        try jws.write("Infinity");
+    } else if (std.math.isNegativeInf(value)) {
+        try jws.write("-Infinity");
+    } else {
+        try jws.write(value);
+    }
+}
+
+fn strinfigy_struct_field(
```
james-callahan, line 1229:
```suggestion
fn stringify_struct_field(
```

src/protobuf.zig
```diff
@@ -963,6 +987,347 @@ pub fn pb_decode(comptime T: type, input: []const u8, allocator: Allocator) !T {
...
+            capitalize_next_letter = false;
+            i += 1;
+        } else {
+            camel_cased_string[i] = char;
+            i += 1;
+        }
+    }
+
+    if (std.ascii.isUpper(camel_cased_string[0])) {
+        camel_cased_string[0] = std.ascii.toLower(camel_cased_string[0]);
```
james-callahan, line 1206: This misses the case where the string starts with `_` then a capital letter; is that on purpose?

src/protobuf.zig
```diff
@@ -980,6 +1345,159 @@ pub fn MessageMixins(comptime Self: type) type {
...
+                        // array will be returned (there's no need
+                        // for deallocation in this case)
+                        yes2 = false;
+                    } else {
+                        yes2 = std.mem.eql(u8, camel_case_name, field_name);
+                        allocator.free(camel_case_name);
+                    }
+
+                    if (yes1 and yes2) {
+                        return error.FieldNameCollision;
```
james-callahan, line 1415: This results in:
```
/usr/lib/zig/std/json/static.zig:322:35: note: 'error.FieldNameCollision' not a member of destination error set
/usr/lib/zig/std/json/static.zig:211:32: note: function return type declared here
) ParseError(@TypeOf(source.*))!T {
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
referenced by:
```



I think you need:
```suggestion
                        return error.DuplicateField;
```
which is in the json error set.

src/protobuf.zig
librolibro: I just took the implementation from [this python code](https://github.com/protocolbuffers/protobuf/blob/06a520cdb78be66f506b88a8adba2763f2301dc0/python/google/protobuf/descriptor.py#L1299) and also [this C++ code](https://github.com/protocolbuffers/protobuf/blob/06a520cdb78be66f506b88a8adba2763f2301dc0/hpb_generator/gen_messages.cc#L126). Since it's the official google libraries others supported languages probably have the same conversion code.

src/protobuf.zig
librolibro: That's the type of confusion I talked about. IMO DuplicateField throws when JSON string contains two entries of one key, but that's the opposite case when one dictionary entry match with two struct fields. It's bad naming anyway and user should avoid it but it's still valid code and because of this specific error he might spend more time finding what's wrong and why his code is not working. That's why I thought to use panic - if you want to work with **bad** structure without JSON conversion everything will be fine, but as soon as you'll try to (de)serialize it via JSON there will a panic "Boy you'll shoot yourself in the leg with this". Is it good to not let user generate structures from these **bad** message definitions?

Maybe it's fine and I'm overreacting :) I think it will be the rare case anyway that someone would choose different name style for message fields using the same words but different *delimiters*.

Arwalk:
> It seems as though the allocator is just used for the camel case functionality; but that only needs to work at compile time.
So you should be able to swap to a comptime implementation:

Excellent remark, we have everything needed to know this at compile time indeed.

> From there it is a quick cleanup, aside from the base64 encoding, which needs to drop down to writing to the json stream directly rather than allocating a temporary.

I'm gonna check if there are ways to do that, but i'm not sure i have access to the slices in the json encoding api.
In the end, it's not dramatic if this stays as is, if you're already working with json encoding/decoding you're probably not too worried about a pair of temporary allocations.

librolibro:
I wanted to commit camelCase making in comptime but It'll be awkward if you're making the same)

src/protobuf.zig
librolibro: In case you're talking about it'll skip all leading underscores and then will make first letter the lowercase one

james-callahan:
> I'm gonna check if there are ways to do that, but i'm not sure i have access to the slices in the json encoding api. In the end, it's not dramatic if this stays as is, if you're already working with json encoding/decoding you're probably not too worried about a pair of temporary allocations.
> 
> Edit: nevermind, found a way in [1c8ab07](https://github.com/Arwalk/zig-protobuf/commit/1c8ab074f7cf66af537091ee78a355199a2110e2) . That require rewriting some of the code inside the json writer, we might have to make a request to make some methods public in the std.

In theory you don't need to: a base64 encoder should be able to work on (multiples of) 3-byte pieces.
So e.g. do:
```zig
  const str = value.getSlice();
  try jws.valueStart();
  try jws.stream.writeByte('\"');
  var temp_out: [5]u8;
  var i=0;
  while (i < str.len) : (i+=3) {
    const s = base64.standard.Encoder.encode(&temp_out, str[i..i+3]);
    try jws.stream.writeAll(s);  
  }
  try jws.stream.writeByte('\"');
  jws.valueDone();
```

Arwalk:
> > I'm gonna check if there are ways to do that, but i'm not sure i have access to the slices in the json encoding api. In the end, it's not dramatic if this stays as is, if you're already working with json encoding/decoding you're probably not too worried about a pair of temporary allocations.
> > Edit: nevermind, found a way in [1c8ab07](https://github.com/Arwalk/zig-protobuf/commit/1c8ab074f7cf66af537091ee78a355199a2110e2) . That require rewriting some of the code inside the json writer, we might have to make a request to make some methods public in the std.
> 
> In theory you don't need to: a base64 encoder should be able to work on (multiples of) 3-byte pieces. So e.g. do:
> 
> ```zig
>   const str = value.getSlice();
>   try jws.valueStart();
>   try jws.stream.writeByte('\"');
>   var temp_out: [5]u8;
>   var i=0;
>   while (i < str.len) : (i+=3) {
>     const s = base64.standard.Encoder.encode(&temp_out, str[i..i+3]);
>     try jws.stream.writeAll(s);  
>   }
>   try jws.stream.writeByte('\"');
>   jws.valueDone();
> ```

Could have worked indeed. I managed to make it in one pass, that should be good enough.

librolibro:
**repeated bytes**, **oneof=NaN, +/-inf** fails as expected (NaN in arrays and oneof=.Bytes would probably fail as well), I'll make some changes to support these, for now there are failing tests to not cover this problem

Arwalk:
> P.P.S. Is there a way to disable workflow runs temporarily? I certainly know that tests will fail because I wrote them to fail :) Or maybe I shouldn't push until I won't make it good

It's actually me that's starting them manually sometimes, just tell me when you feel they should run.

librolibro:
For now it looks like all (supported ATM) types are processed as expected. I think the current JSON code base is messy IMO - especially in case of function naming. I think it's better to refactor all JSON code according to your code style and not to Zig's one (snake_case for functions instead of camelCase etc.) to avoid more refactor for now. Maybe it's time to move JSON code to separate file because there's a lot of things copied from **std.json** + things implemented by use (400+ lines of code).

Also more tests ... For things like deserializing integer/float from JSON string (whether it's single, repeated of inside of oneof) etc.

librolibro:
Please run another workflow for this PR. I want to check if tests will pass now of Windows (there were strange fails, it was probably about different line endings, but I'm not 100% sure)

Arwalk:
Looks like it's not fixed yet. Don't bother too much, i'll take the time to look into it tuesday. Thank you for all your work already.

librolibro:
Now .gitattributes work as expected (for some reason leading **./** was the problem and I missed it). If Windows build failures are about line endings it might fix the problem now :)

```bash
$ git check-attr --all -- tests/json_data/fixed_sizes/camelCase_1.json
tests/json_data/fixed_sizes/camelCase_1.json: text: set
tests/json_data/fixed_sizes/camelCase_1.json: eol: lf
```

librolibro:
Last major thing wasn't made yet is `Always emit fields without presence` described [here](https://protobuf.dev/programming-guides/proto3/#json-options) + tests for cases when not all fields are present in the input JSON string (I think there's MessageMixins.init call missed and these tests would fail).

For now I don't know how to vary parser's behavior (omit default fields or not, serialize using camelCase or snake_case etc.), there's no able to pass extra state into std.json inner functions and hence into our custom jsonParse/jsonStringify implementations. I have couple "possibly good" ideas how to make it, I would try these after default fields workaround, maybe in the weekend.

Arwalk (approved):

Arwalk:
This will be good enough for a good "basic" json encoding/decoding. I'll create issues for everything mentionned in https://github.com/Arwalk/zig-protobuf/pull/49#issuecomment-2249600720 So we can do bit by bit everything right.

Thanks a lot for your involvment !

Arwalk:
@librolibro the following issues have been created, so PR can be done to target one subject at a time:

- #59 
- #60 
- #61 
- #62

librolibro:
I was on the way to push some commits :) Earlier I thought that Github PR's can be reopened with merge up-to-date master with fork's branch, and then do it again for next issue (does Github allows this?)... But separate issues/PRs might be a cleaner solution.

Arwalk:
Woops sorry! I'd really prefer if we split it up, it will be cleaner to handle.

PR #58, opened 2024-08-06
WIP: Add support for protos without package name

osztenkurden:
There is still formatting to do, but beside that it works. One thing that needs to be taken care of, is that `import "networkbasetypes.proto";` does not generate proper `@import` clause

Also haven't tested if it breaks something, hence very much WIP


Arwalk:
Hey there @osztenkurden , where are you on this? Do you have any plan? Please do not force yourself to finish this, if you want to leave the subject it's your call and would be reasonable. Thanks again for your work and interest already.

osztenkurden:
Hey, I've been on vacations for 2 weeks, I shall be back with it soon. Ill try to have it done next week, as I also need it for my project ;D

Issue #55, opened 2024-07-29
How to use system/different `protoc`


james-callahan:
From a supply chain security point of view I don't want zig-protobuf to provide its own protoc: I'd like to either use my system one, or possibly build it from source.
How can I tell zig-protobuf to use my system protoc?

Arwalk:
Fair point, we might need to add options to RunProtoc. At first glance I'd say we need 

- download (current behaviour, default)
- use system provided through path encmv
- use user provided protoc

PR #66, opened 2024-08-12
Allow for external protoc

BonusPlay:
This PR adds partial support for #55 via `PROTOC_PATH` environment variable.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -287,6 +287,20 @@ fn getProtocInstallDir(
     allocator: std.mem.Allocator,
     protoc_version: []const u8,
 ) ![]const u8 {
+    if (std.process.getEnvVarOwned(allocator, "PROTOC_PATH")) |protoc_path| {
+        std.log.info("zig-protobuf: respecting PROTOC_PATH: {s}\n", .{ protoc_path });
+        if (fileExists(protoc_path)) {
+            // user has probably provided full path to protoc binary instead of proto_dir
+            // also, if these fail and user explicitly provided custom path, we probably don't want to download stuff
+            const bin_dir = std.fs.path.dirname(protoc_path).?;
+            const real_proto_dir = std.fs.path.dirname(bin_dir).?;
+            return real_proto_dir;
+        }
+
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -287,6 +287,20 @@ fn getProtocInstallDir(
...
+            // user has probably provided full path to protoc binary instead of proto_dir
+            // also, if these fail and user explicitly provided custom path, we probably don't want to download stuff
+            const bin_dir = std.fs.path.dirname(protoc_path).?;
+            const real_proto_dir = std.fs.path.dirname(bin_dir).?;
+            return real_proto_dir;
+        }
+
+        std.log.err("zig-protobuf: cannot resolve a protoc provided via PROTOC_PATH env var ({s}), make sure the value is correct", .{protoc_path});
+        std.process.exit(1);
+    } else |_| {}
```
Arwalk, line 302: You can use `catch null` to avoid an empty error capture into an empty block in else.

```suggestion
    if (std.process.getEnvVarOwned(allocator, "PROTOC_PATH") catch null) |protoc_path| {
        std.log.info("zig-protobuf: respecting PROTOC_PATH: {s}\n", .{ protoc_path });
        if (fileExists(protoc_path)) {
            // user has probably provided full path to protoc binary instead of proto_dir
            // also, if these fail and user explicitly provided custom path, we probably don't want to download stuff
            const bin_dir = std.fs.path.dirname(protoc_path).?;
            const real_proto_dir = std.fs.path.dirname(bin_dir).?;
            return real_proto_dir;
        }

        std.log.err("zig-protobuf: cannot resolve a protoc provided via PROTOC_PATH env var ({s}), make sure the value is correct", .{protoc_path});
        std.process.exit(1);
    }
```

Arwalk (changes requested):
Apart from a very minor change, this works. Thank you.

BonusPlay:
I've tested it on nixos, which wouldn't work without this PR.

Arwalk (approved):

Arwalk:
I tested it on the devcontainer too and it worked fine.

Arwalk:
Thanks again!

PR #68, opened 2024-09-14
fix: memory leak when decoding Lists of Bytes

kantai-chris:
Messages that include a `repeated` field of type `bytes` are not properly freed due to a missing condition in `deinit_field()` which only cares for list of other message types or strings, but not bytes.

This commit introduces a new unit test that fails if decoding a message with a list of bytes leaks and fix, that makes the unit test pass.

### DIFF
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -586,7 +586,7 @@ fn deinit_field(result: anytype, comptime field_name: []const u8, comptime ftype
             }
         },
         .List => |list_type| {
-            if (list_type == .SubMessage or list_type == .String) {
+            if (list_type == .SubMessage or list_type == .String or list_type == .Bytes) {
                 for (@field(result, field_name).items) |item| {
                     item.deinit();
                 }
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -586,10 +586,13 @@ fn deinit_field(result: anytype, comptime field_name: []const u8, comptime ftype
             }
         },
         .List => |list_type| {
-            if (list_type == .SubMessage or list_type == .String or list_type == .Bytes) {
-                for (@field(result, field_name).items) |item| {
-                    item.deinit();
-                }
+            switch (list_type) {
+                .SubMessage, .String, .Bytes => {
+                    for (@field(result, field_name).items) |item| {
+                        item.deinit();
+                    }
+                },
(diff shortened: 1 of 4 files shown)


Arwalk:
Thanks! I'll look into the details ASAP. It seems I introduced this bug myself recently with bytes. Seems OK at first glance, but I will look at it properly soon.

src/protobuf.zig
```diff
@@ -586,7 +586,7 @@ fn deinit_field(result: anytype, comptime field_name: []const u8, comptime ftype
             }
         },
         .List => |list_type| {
-            if (list_type == .SubMessage or list_type == .String) {
+            if (list_type == .SubMessage or list_type == .String or list_type == .Bytes) {
```
Arwalk, line 589: Let's use switch here so we have the exhaustive check at compile time and avoid this kind of error in the future.

```patch
diff --git a/src/protobuf.zig b/src/protobuf.zig
index dbe0c48..ac7e811 100644
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -586,10 +586,13 @@ fn deinit_field(result: anytype, comptime field_name: []const u8, comptime ftype
             }
         },
         .List => |list_type| {
-            if (list_type == .SubMessage or list_type == .String or list_type == .Bytes) {
-                for (@field(result, field_name).items) |item| {
-                    item.deinit();
-                }
+            switch (list_type) {
+                .SubMessage, .String, .Bytes => {
+                    for (@field(result, field_name).items) |item| {
+                        item.deinit();
+                    }
+                },
+                .Varint, .FixedInt => {},
             }
             @field(result, field_name).deinit();
         },
```

Arwalk (changes requested):
Just a slight little change to enforce switch (which would have prevented the mistake in the first place).

But apart from that this is great, thanks for the catch!

src/protobuf.zig
kantai-chris: Ok, I'll change it. Thanks for the improvement. I'm currently taking my first baby steps with zig, so I'm grateful for each learning along the way :)

kantai-chris:
I changed the check to use `switch` instead of `if`, please review.

Arwalk (approved):

Arwalk:
Thanks a lot for your contribution !

PR #70, opened 2024-10-11
fix : breaking changes on zig 0.14

AlpaGit:
std.builtin.Type has changed and is now using lower case naming convention
build.zig changed some methods but nothing crazy

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -192,8 +192,8 @@ pub const RunProtocStep = struct {
         self.step.name = name;
     }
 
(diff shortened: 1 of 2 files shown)


AlpaGit:
my bad, it was already done in
https://github.com/Arwalk/zig-protobuf/commit/bcd11ef0ee4e490bf26caadf5c78845dc0207b21

Arwalk:
I checked that zig-master is compatible with the current latest version (0.14.0-dev.1860+2e2927735), that should be fine. I'm closing this.

Thanks for your interest.

PR #71, opened 2024-10-14
Fix: prevent unsafe cast on decoding

AlpaGit:
I ran into an issue that would cause my program to crash when trying to decode an incorrect input, this would at least prevent some kind of errors, I also added a test to replicate the behavior


Arwalk:
Hello @AlpaGit , thanks a lot! Could you please do the fix on the `master` branch so we can then port it on `zig-master` ?
If you don't wanna do it it's fine, i'll do it myself later tonight.

@menduz thanks for the review.

Arwalk:
Ok, doing it myself right now. Thanks a lot, you will be credited for your contribution @AlpaGit .

Sorry for being annoying, i really want to keep the approach of developping on master then merging to zig-master.

PR #73, opened 2024-10-14
Draft: 69 Self referential types in `oneOf` results in error

### LINKED ISSUE #69: Self referential types in `oneOf` results in error
If there are self referential types in `oneOf`, then it results in the following error from Zig: `struct 'proto.foobar.pb.Node' depends on itself`. 

Here's an example proto file:

```proto
syntax = "proto3";

package foobar;

message Result {
  int32 version = 1;
  Node node = 2;
}

message Node {
  oneof node {
    SubNode sub_node = 1;
    string some_string = 2;
  }
}

message SubNode {
  Node sub = 1;
  string another_string = 2;
}
```

The corresponding main file:

```
const std = @import("std");
const proto = @import("./proto/foobar.pb.zig");

pub fn main() !void {
    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
    defer _ = gpa.deinit();

    var arena = std.heap.ArenaAllocator.init(gpa.allocator());
    defer arena.deinit();

    const foo = try protobuf.pb_decode(proto.Result, "", arena.allocator());
    _ = foo;
}
```

Arwalk:
closes #69


src/protobuf.zig
```diff
@@ -100,6 +108,45 @@ pub const ManagedString = union(ManagedStringTag) {
...
+        Owned: AllocatedStruct(T),
+        Managed: *const T,
+
+        const Self = @This();
+
+        pub fn managed(p : *const T) ManagedStruct(T){
+            return ManagedStruct(T) {.Managed = p};
+        }
+
+        pub fn move(p: *T, allocator : Allocator) ManagedStruct(T) {
```
menduz, line 127: Probably a doc comment about who owns the memory and should be responsible for the deallocation would be nice here

menduz (commented):
🙌

tests/generated/selfref.pb.zig
```diff
@@ -0,0 +1,21 @@
...
+const Allocator = std.mem.Allocator;
+const ArrayList = std.ArrayList;
+
+const protobuf = @import("protobuf");
+const ManagedString = protobuf.ManagedString;
+const fd = protobuf.fd;
+
+pub const SelfRefNode = struct {
+    version: i32 = 0,
+    node: ?SelfRefNode = null,
```
menduz, line 13: I think we could make this much easier if we used a pointer for self referencial nodes. Or if we always used a pointer.


Although the new recommendation for protobuf implementations, is to use moving or copying getters and setters in addition to opaque structs. Enabling internal reordering and packing

Arwalk: I feel like we would lose some quality of use if we were to use a pointer everywhere. SelfRef structures are not uncommon, but i feel like they're far from being a common use case. I'd like to keep the common use case as simple and straightforward as possible.

My goal in this branch is to identify cases where there is some self ref structures in the structure hierarchy when running the plugin, and only using ManagedStruct in these case.

Arwalk:
Disclaimer, the last commit was entirely done using AI, i was experimenting with it. The entire session is available here: https://gist.github.com/Arwalk/e20038ee27b08f0c5df8c95291943612

src/protobuf.zig
```diff
@@ -1043,25 +1149,57 @@ inline fn is_tag_known(comptime field_desc: FieldDescriptor, tag_to_check: Extra
...
+        }
+
+        pub fn getTarget(self: *Self) *rootType {
+            return if (isZigProtobufManagedStruct(T))
+                self.result.getPointer()
+            else
+                return &self.result;
+        }
+    };
+}
```
menduz, line 1180: I think we can convert this struct to two functions with comptime parameters. AFAIK the struct is not being passed and only holds a value which can 1) be returned in `init` and 2) be passed to getTarget

src/protobuf.zig
```diff
@@ -449,21 +533,38 @@ fn append(pb: *ArrayList(u8), comptime field: FieldDescriptor, value: anytype, c
...
+                data;
+
+            try internal_append(pb, data_type, field.name, value, false);
         }
     }
 }
 
+fn internal_append(pb: *ArrayList(u8), data_type: type, comptime field_name: []const u8, value: anytype, comptime force_append: bool) !void {
+    try append(pb, @field(data_type._desc_table, field_name), @field(value, field_name), force_append);
+}
```
menduz, line 566: maybe we can inline this one?

src/protobuf.zig
```diff
@@ -449,21 +533,38 @@ fn append(pb: *ArrayList(u8), comptime field: FieldDescriptor, value: anytype, c
...
-                if (@field(data, field.name)) |value| {
-                    try append(pb, @field(data_type._desc_table, field.name), value, true);
-                }
-            },
-            else => try append(pb, @field(data_type._desc_table, field.name), @field(data, field.name), false),
+        if (@typeInfo(field.type) == .Optional) {
+            const temp = if (isZigProtobufManagedStruct(@TypeOf(data)))
+                data.get()
+            else
+                data;
```
menduz, line 549: Since it is a very repeating pattern, I think there is a trivial way to write this if as the suggestion below resulting in the same final compiled code.
```suggestion
            const temp = getPointerOrValue(data);
```

src/protobuf.zig
```diff
@@ -23,6 +23,27 @@ const UnionDecodingError = DecodingError || Allocator.Error;
 
 pub const ManagedStringTag = enum { Owned, Const, Empty };
 
+/// This structure is used by ManagedStruct to hold a T allocated.
+fn AllocatedStruct(T: type) type {
```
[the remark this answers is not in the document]
menduz: The only drawback being the convenience helpers like `getSlice`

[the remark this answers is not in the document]
Arwalk: I thought about trying to have ManagedString being just an instance of ManagedStruct, but i think this would be factorisation for the sake of factorisation, with an added cost on the user's side. ".Empty" does not have a lot of sense for structures for example, when i think it's a convenient information.

We can talk about it in another issue, if one of us (or somemone else) wants to experiment on this subject. Ok ?

src/protobuf.zig
```diff
@@ -100,6 +121,69 @@ pub const ManagedString = union(ManagedStringTag) {
...
+pub const ManagedStructTag = enum { Owned, Borrowed };
+
+pub fn ManagedStruct(T: type) type {
+    return union(ManagedStructTag) {
+        const Self = @This();
+        const UnderlyingType = T;
+        const AllocatedType = AllocatedStruct(T);
+
+        Owned: AllocatedType,
+        Borrowed: *T,
```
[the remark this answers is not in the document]
Arwalk: > Should we include Empty and remove all nullable optionals from the code?

i'd prefer we rely on the language's intrinsic capabilities.

> Should we include Value: T and remove all necessity for pointers for a % of cases?

isn't that why `get` is here ?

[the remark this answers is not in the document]
menduz: > i'd prefer we rely on the language's intrinsic capabilities.

Thats reasonable, but it is not in sync with how we handle strings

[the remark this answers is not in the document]
Arwalk: > Thats reasonable, but it is not in sync with how we handle strings

Yeah, strings being just u8 arrays makes us do an exception. So let me rephrase this: "I'd prefer we rely on the language's intrinsic capabilities as much as possible. Because we choose not to use pointers everywhere, we are kinda forced to have a few exceptions around".

Using pointers everywhere would solve all of our problem, but as someone who myself uses protobuf in restricted environments such as embedded  (sadly not with zig), i see a lot of value in our approach.

src/protobuf.zig
```diff
@@ -100,6 +127,73 @@ pub const ManagedString = union(ManagedStringTag) {
...
+
+        /// This const's only purpose is to identify properly ManagedStructs.
+        const isZigProtobufManagedStruct = true;
+
+        /// Creates a ManagedStruct.Borrowed instance that has a pointer to an existing T, but do not take ownership of its memory.
+        ///
+        /// This means that a call to deinit on this instance will not free the memory associated with it. The original T's deinit() function must be called for that.
```
menduz, line 149: I had to read many times to understand, if you agree on my comment would you rephrase it?

I'm seeing the pattern in many zig codebases being explicit about "caller owns the memory" to signal who should deinit the thing.

menduz:
Implementation looks good so far, I'd like to spend a bit of time iterating on comments and names (as changing the names down the road may be a breaking change best avoided)

Arwalk:
> Implementation looks good so far, I'd like to spend a bit of time iterating on comments and names (as changing the names down the road may be a breaking change best avoided)

Ok, waiting for your suggestions on comments. We should avoid making some breaking changes for now. We have a subject about aligning the whole codebase on zig's std style, but this will be a breaking change so we'll do it after 0.14.0 release.

src/protobuf.zig
```diff
@@ -62,6 +87,7 @@ pub const ManagedString = union(ManagedStringTag) {
         }
     }
 
+    /// Creates a deep copy of the managed string using the provided allocator
```
menduz, line 90:
```suggestion
    /// Creates a copy of the managed string using the provided allocator
```
There are not shallow or deep copies in []u8

src/protobuf.zig
menduz: After a couple days of thinking, this is my suggestions for the public facing API.

Here is a bullet list of the changes detailed below as pseudo-diffs:
- Each message will now store a reference to its allocator (set in init(), decode() and dupe()). Making it "managed" (in zig slang)
- Each message now exposes helper functions to create references (via mixins)
  - `fn move(self): ManagedStruct(Self)` and
  - `fn reference(self): ManagedStruct(Self)`


```diff
pub fn MessageMixins(comptime Self: type) type {
    return struct {
+       /// each message will store a pointer to its allocator. Value is set in init(allocator)
+       allocator: Allocator,

+       /// Creates a ManagedStruct for this message. 
+       /// Moves the ownership of the memory to the receiver of the ManagedStruct.
+       pub fn move(ref: *@This()): ManagedStruct(Self) {
+         return .{ .Managed = ref };
+       }

+       /// Creates a ManagedStruct for this message. 
+       /// Caller owns the memory.
+       pub fn reference(ref: *@This()): ManagedStruct(Self) {
+         return .{ .Unmanaged = ref };
+       }
    }
}
```

Since each message references its allocator, it is is now "managed" in zig idiom.

1. With that assumption, a lot of complexity from ManagedStruct can be removed down to a single pointer + memory ownership tag.
2. `ManagedStruct` is redundant name, but we still need to be explicit about the kind of usage and behaviors of ownership.

```zig
pub fn ManagedStruct(T: type) type {
    return union(ManagedStructTag) {
        Owned: *T,
        Borrowed: *T,
    }
}
```

Maybe `MessagePointer` is a better name for the reduced scope:

```diff
-pub fn ManagedStruct(T: type) type {
+pub fn MessagePointer(T: type) type {
    return union(ManagedStructTag) {
        Owned: *T,
        Borrowed: *T,
    }
}
```

```zig

test "self ref test, moving semantics" {
    var demo = SelfRefNode.init(testing.allocator);
    var demo2 = SelfRefNode.init(testing.allocator);
    demo.node = SelfRefNode.move(demo2);
    defer demo.deinit();
    // TODO: assert demo2 was deinitialized by demo.deinit()
}

test "self ref test, moving semantics" {
    var demo = SelfRefNode.init(testing.allocator);
    var demo2 = SelfRefNode.init(testing.allocator);
    demo.node = SelfRefNode.reference(demo2);
    defer demo.deinit();
    // TODO: assert demo2 is was NOT deinitialized
}

```




---


Open questions: 
- what will happen with dupe() with all this managed/unmanaged pointers?
  ```zig

  test "dupe" {
      var demo1 = SelfRefNode.init(testing.allocator);
      var demo2 = SelfRefNode.init(testing.allocator);
      demo1.node = SelfRefNode.move(demo2);
      var demo1dupe = demo1.dupe(testing.allocator);

      // WHO OWNS demo1dupe.node??????
  }
  ```
- should we add a safeguard to prevent trying to serialize a self-referencial value?
  ```zig

  test "prevent stack overlow" {
      var demo = SelfRefNode.init(testing.allocator);
      demo.node = SelfRefNode.move(demo);
      // SHOULD THIS FAIL WITH A BETTER ERROR THAN STACK OVERFLOW/OOM?
      try demo.encode(..)
  }
  ```

menduz:
Here are my final comments https://github.com/Arwalk/zig-protobuf/pull/73#discussion_r1976647325

menduz (approved):

src/protobuf.zig
Arwalk: Ok, approach seems fine and prob, and as long as we put these only when there are self-ref nodes, it should be fine. I'll take a look into it when i can.

> what will happen with dupe() with all this managed/unmanaged pointers?

In your example, demo1.dupe should be duplicated using the allocator that was passed in `demo1.dupe(testing.allocator)`. That's the only "explicit" choice in the movement. In the end, thats the kind of thing we can only just pick a choice and stand our ground on it.

> should we add a safeguard to prevent trying to serialize a self-referencial value?

Ideally, yes, but maybe i'll leave it for later

Arwalk: Yeah, i've played with it again and it makes everything a hassle for any struct that doesn't require an allocator. Making it optional doesn't help and just makes more work for nothing.

I really think we have a good interface here and transforming everything just for the case of self referential structures seems like a bad idea too me.

menduz: :shipit:

PR #76, opened 2024-10-28
fix: Convert OOB enums from panics to runtime errors

hans-tvs:


Arwalk (changes requested):
Excellent! Some minor changes requested. Thank you for catching this problem and fixing it.

src/protobuf.zig
```diff
@@ -825,7 +825,10 @@ fn decode_varint_value(comptime T: type, comptime varint_type: VarintType, raw:
                 else => @compileError("Invalid type " ++ @typeName(T) ++ " passed"),
             },
             .Bool => raw != 0,
-            .Enum => @as(T, @enumFromInt(@as(i32, @intCast(raw)))),
+            .Enum => if ((raw >> 32) != 0)
+                error.InvalidInput
+            else
+                @as(T, @enumFromInt(@as(i32, @bitCast(@as(u32, @intCast(raw)))))),
```
menduz, line 831: Why is there a new bitCast?

menduz (commented):
Could you add some comments about the motivations for the changeset? the motivation is not evident in the tests

hans-tvs:
Thanks for the suggestions. I took another pass at this.

> explicit error type

I used that particular error literal because it fit the problem well and appeared many times throughout the code. Does re-using the `DecodingError` you already have defined work here?

> `std.math.maxInt(i32)`
> Why is there a new bitcast?

These are related. It's valid for protobuf enums to have negative values, and the previous `@intCast` didn't suffice for those. The solution here uses an `@intCast`, as before`, to drop from 64 to 32 bits (cost varies from architecture to architecture, but it's cheap), and a `@bitCast` to assert that those bits are actually signed (free at runtime except in some esoteric microcontrollers and whatnot).

Regarding your suggestion, comparing to `std.math.maxInt(u32)` (not the signed version) seems reasonable.

> Can you add some comments about the motivations for the changeset?

I added comments in code for one of the tests; the other test was just a new test for existing behavior (mainly for my benefit, ensuring I hadn't made any egregious errors, but it usually doesn't hurt to add those tests upstream). I'm not totally sure what you're looking for, so would you mind taking a peek and explaining why that comment does or does not work?

menduz (approved):
Many thanks for the revisions

Arwalk (approved):

Arwalk:
Thanks again for your contribution.

hans-tvs:
Thank you so much for the fast review and for building this project in the first place :)

PR #79, opened 2024-10-30
update for zig 0.14.0-dev.2064+b5cafe223

mindon:


Arwalk:
Hello there, thanks for your contribution. As stated in the readme, the master branch is for stable zig, while the zig-master tries to be compatible with the latest zig. 

Can you try zig-master, and if any change is needed make a pr for this branch?

Thanks a lot.

PR #86, opened 2025-01-13
Fix integer encoding

Arwalk:
This branch aims at fixing i32 encoding as identified in https://github.com/Arwalk/zig-protobuf/discussions/82
I also took the opportunity to have a better handling of most zig zag encoding, as i identified that it was breaking on limits.

This will be a major release as it is technically a breaking change. We will have to cover what the error was and how to recover from it if it was encountered in production.

@menduz sorry for the burden but it would be nice to have a thorough review, i generated the values for the tests with some other reference implementations, but if you could do it on your side too... Thanks in advance.

### DIFF
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -194,17 +194,42 @@ fn insert_raw_varint(pb: *ArrayList(u8), size: u64, start_index: usize) Allocato
     }
 }
 
+fn encode_zig_zag(int: anytype) u64 {
+    const type_of_val = @TypeOf(int);
+    const to_int64 : i64 = switch (type_of_val) {
+        i32 => @intCast(int),
+        i64 => int,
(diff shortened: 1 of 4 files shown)


Arwalk:
Hey @menduz , if you don't have the time to review this it's perfectly fine. Should l merge as is or would you still like to look at it? Sorry to bother you.

menduz (approved):

Arwalk:
> Sorry for the delay, I'm traveling and I hadn't had the time to dig deeper into this.
> 
> The change looks sensible and it passes more tests than before. We should be able to publish a new version and it will be definitely better than before. Maybe a major and an advisory warning will be required in the readme.
> 
> I only hope there are not more breaking changes like this one 🤞 

No need to apologise, take care, and thank you.

PR #90, opened 2025-01-30
Explicit error types

### LINKED ISSUE #77: Making all error types explicits instead of inferred
I realize that in all our methods, we only use the inferred error type. This is overall not a problem, but when i see that #74 happened, this makes me think that it would be a nice improvement to explicitely set the errors in our method's interface.

Also, i feel like any serious error handling code would, for example, check if it's an invalid input or an allocation error, and recover if necessary.

the json codec would be optional on this subject, i'm not sure if it's easily feasible to keep it explicit.

42LM:
Hello there 👋.

This PR adds explicit error returns everywhere where it is easily feasible.

Methods/Funcs that return a e.g. `json.ParseError(@TypeOf(source.*))` keep the inferred error return type.

Methods/Funcs that use the writer interface in form of using e.g. `.write()` could return `anyerror`, but i do not think that this is helpful. (For more context: [anyerror-void-versus-void](https://ziggit.dev/t/anyerror-void-versus-void/5567)

Have a great weekend everyone.

> resolves #77

### DIFF
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -743,7 +743,7 @@ fn VarintDecoderIterator(comptime T: type, comptime varint_type: VarintType) typ
         input: []const u8,
         current_index: usize = 0,
 
-        fn next(self: *Self) !?T {
+        fn next(self: *Self) DecodingError!?T {
             if (self.current_index < self.input.len) {
                 const raw_value = try decode_varint(u64, self.input[self.current_index..]);
                 defer self.current_index += raw_value.size;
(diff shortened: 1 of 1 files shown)


src/protobuf.zig
```diff
@@ -1072,7 +1072,7 @@ fn fillDefaultStructValues(
     comptime T: type,
     r: *T,
     fields_seen: *[@typeInfo(T).Struct.fields.len]bool,
-) !void {
+) error{MissingField}!void {
```
Arwalk, line 1075: This is part of the json stuff, so not really needed, but it's ok.

Arwalk (approved):
Appart from things that are related to json and subject to change in a not so far future, everything has now an explicit error type (when needed), which is great.

src/protobuf.zig
42LM: Nice! Yes it was kind of a low hanging fruit :)

PR #91, opened 2025-01-31
Switch remove else clause

### LINKED ISSUE #78: Remove all `else` clauses when possible
#68 was a direct consequence of not using exhaustive option checks when possible. We should rely on the compiler to tell us that something's wrong if possible, and using `switch` is a good way to do it.

Edit: a word

42LM:
Hey there,
this PR exchanges else prongs with explicit types using exhaustive option checks when possible.

It only exchanges where it seemed reasonable to me. So there is still room for more exchanges here like e.g.:
```zig
fn append_varint(pb: *ArrayList(u8), value: anytype, comptime varint_type: VarintType) Allocator.Error!void {
    switch (@typeInfo(@TypeOf(value))) {
        .Enum => try append_as_varint(pb, @as(i32, @intFromEnum(value)), varint_type),
        .Bool => try append_as_varint(pb, @as(u8, if (value) 1 else 0), varint_type),
        else => try append_as_varint(pb, value, varint_type),
    }
}
```

Since the `std.builtin.Type` is a pretty large union the list gets pretty big and i did not like it:
```zig
fn append_varint(pb: *ArrayList(u8), value: anytype, comptime varint_type: VarintType) Allocator.Error!void {
    switch (@typeInfo(@TypeOf(value))) {
        .Enum => try append_as_varint(pb, @as(i32, @intFromEnum(value)), varint_type),
        .Bool => try append_as_varint(pb, @as(u8, if (value) 1 else 0), varint_type),
        .Type, .NoReturn, .Int, .Float, .Pointer, .Array, .Struct, .ComptimeFloat, .ComptimeInit, .Undefined, .Null, .Optional, .ErrorUnion, .ErrorSet, .Union, .Fn, .Opaque, .Frame, .AnyFramge, .Vector, .EnumLiteral => try append_as_varint(pb, value, varint_type),
    }
}
```

While i am more okey with this last one, it also consumes quite an amount of lines plus `zls` always reformats it to the above example ;)
```zig
fn append_varint(pb: *ArrayList(u8), value: anytype, comptime varint_type: VarintType) Allocator.Error!void {
    switch (@typeInfo(@TypeOf(value))) {
        .Enum => try append_as_varint(pb, @as(i32, @intFromEnum(value)), varint_type),
        .Bool => try append_as_varint(pb, @as(u8, if (value) 1 else 0), varint_type),
        .Type,
        .NoReturn, 
        .Int, 
        .Float, 
        .Pointer, 
        .Array,
        .Struct, 
        .ComptimeFloat,
        .ComptimeInit, 
        .Undefined, 
        .Null, 
        .Optional, 
        .ErrorUnion, 
        .ErrorSet, 
        .Union, 
        .Fn, 
        .Opaque, 
        .Frame, 
        .AnyFramge, 
        .Vector, 
        .EnumLiteral => try append_as_varint(pb, value, varint_type),
    }
}
```

Is there something one could do about this? My zig knowledge comes to an end here, in my dreams i would somehow put those values in a const and than use that in the switch. However i was not successful with this.

WDYT? Should the `else` prong stay in the case of a huge list?

> resolves #78

### DIFF
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -535,14 +535,17 @@ fn dupe_field(original: anytype, comptime field_name: []const u8, comptime ftype
         .List => |list_type| {
             const capacity = @field(original, field_name).items.len;
             var list = try @TypeOf(@field(original, field_name)).initCapacity(allocator, capacity);
-            if (list_type == .SubMessage or list_type == .String) {
-                for (@field(original, field_name).items) |item| {
-                    try list.append(try item.dupe(allocator));
-                }
-            } else {
-                for (@field(original, field_name).items) |item| {
-                    try list.append(item);
-                }
+            switch (list_type) {
+                .SubMessage, .String => {
+                    for (@field(original, field_name).items) |item| {
+                        try list.append(try item.dupe(allocator));
+                    }
+                },
+                .Varint, .Bytes, .FixedInt => {
+                    for (@field(original, field_name).items) |item| {
+                        try list.append(item);
+                    }
+                },
             }
             return list;
         },
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -78,7 +78,7 @@ pub const ManagedString = union(ManagedStringTag) {
             .Owned => |alloc_str| {
                 alloc_str.allocator.free(alloc_str.str);
             },
-            else => {},
+            .Const, .Empty => {},
         }
     }
 
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -452,12 +452,13 @@ fn internal_pb_encode(pb: *ArrayList(u8), data: anytype) Allocator.Error!void {
     const data_type = @TypeOf(data);
 
     inline for (field_list) |field| {
-        if (@typeInfo(field.type) == .Optional) {
-            if (@field(data, field.name)) |value| {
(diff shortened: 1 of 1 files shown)


Arwalk (changes requested):
I didn't found any other else clause that could be expanded, so this looks OK.
That said, there is a slight change that can be made for the `append_varint` function.

```patch
diff --git a/src/protobuf.zig b/src/protobuf.zig
index 27e5ad6..89168c5 100644
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -249,7 +249,8 @@ fn append_varint(pb: *ArrayList(u8), value: anytype, comptime varint_type: Varin
     switch (@typeInfo(@TypeOf(value))) {
         .Enum => try append_as_varint(pb, @as(i32, @intFromEnum(value)), varint_type),
         .Bool => try append_as_varint(pb, @as(u8, if (value) 1 else 0), varint_type),
-        else => try append_as_varint(pb, value, varint_type),
+        .Int => try append_as_varint(pb, value, varint_type),
+        else => @compileError("Should not pass a value of type " ++ @typeInfo(@TypeOf(value)) ++ "here")
     }
 }
``` 

This should be a better approach for this case than just putting the else method.
Otherwise, it's nice. Thanks.

PR #99, opened 2025-04-06
Fix message when missing unzip dependency

randall-fulton:
Building without `unzip` installed points the user towards a different dependency as the reason for build failure. It's not too difficult to figure out what the actual issue is, but this should make the actual failure cause more clear.


Arwalk (approved):
Thanks for the catch.

PR #100, opened 2025-04-06
98 Protobuf Generation Step errors on Windows due to missing curl and unzip

### LINKED ISSUE #98: Protobuf Generation Step errors on Windows due to missing curl and unzip
first i wanna say nice project! I tried generating the protobuf definition through a build step that was described in the readme.md:
```

pub fn build(b: *std.Build) !void {
    // first create a build for the dependency
    const protobuf_dep = b.dependency("protobuf", .{
        .target = target,
        .optimize = optimize,
    });
    
    ...

    const gen_proto = b.step("gen-proto", "generates zig files from protocol buffer definitions");

        // out directory for the generated zig files
        .destination_directory = b.path("src/proto"),
        .source_files = &.{
            "protocol/all.proto",
        },
        .include_directories = &.{},
    });

    gen_proto.dependOn(&protoc_step.step);
}
````
Unfortunately I got an error saying curl was missing, after installing that it gave me an error that unzip was missing.
I failed to find a version of unzip that works on windows so i gave up and tried it in WSL which had unzip and curl already installed.

I was thinking that relying on OS specific software compromises portability in some sense (could one just compile something like curl or unzip in a zig step?).
Anyhow, it would be nice to mention the dependencies in the readme for oblivious windows users like me :)
have a nice day!

Arwalk:
closes #98

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -288,7 +288,7 @@ fn getProtocInstallDir(
     protoc_version: []const u8,
 ) ![]const u8 {
     if (std.process.getEnvVarOwned(allocator, "PROTOC_PATH") catch null) |protoc_path| {
-        std.log.info("zig-protobuf: respecting PROTOC_PATH: {s}\n", .{ protoc_path });
+        std.log.info("zig-protobuf: respecting PROTOC_PATH: {s}\n", .{protoc_path});
         if (fileExists(protoc_path)) {
             // user has probably provided full path to protoc binary instead of proto_dir
             // also, if these fail and user explicitly provided custom path, we probably don't want to download stuff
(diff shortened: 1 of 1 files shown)


Arwalk:
Hey @nat3Github, do you think you could check if this works on your side too?

nat3Github:
> Hey @nat3Github, do you think you could check if this works on your side too?  

sure thing...

yes it works and thank you very much, very convenient with it working out of the box!
tested on MSW 10, Amd Ryzen 7 4800H (x86)

jlucaso1:
@Arwalk 

This PR's changes, while fixing Windows compatibility, have introduced a regression.

 The `cwd` parameter in `std.process.Child.run()` calls causes `curl --version` and `unzip` commands to fail during the protoc generation step.

**Error:** `zig-protobuf: error: 'curl --version' failed. Is curl not installed?`

**fix that worked to me:** Removing the `cwd` parameter from the `Child.run()` and `Child.init()` calls solve the problem to me.

my env: Arch Linux, Zig 0.15.1

Arwalk:
> @Arwalk
> 
> This PR's changes, while fixing Windows compatibility, have introduced a regression.
> 
> The `cwd` parameter in `std.process.Child.run()` calls causes `curl --version` and `unzip` commands to fail during the protoc generation step.
> 
> **Error:** `zig-protobuf: error: 'curl --version' failed. Is curl not installed?`
> 
> **fix that worked to me:** Removing the `cwd` parameter from the `Child.run()` and `Child.init()` calls solve the problem to me.
> 
> my env: Arch Linux, Zig 0.15.1

Hello jlucaso, I see that #138 is ongoing and should solve the problem once it's done. Thank you for your feedback.

PR #102, opened 2025-06-18
Type name includes the outermost type

aaumar25:
Recently, I have a problem when having the following example proto file

```
syntax = "proto3";
package dummy;

message A{ // This is a top-level message named 'A'
  int32 az = 1;
  int32 ay = 2;
}

message B{
  // This is a *nested* message also named 'A'
  // It is scoped *only* within message B.
  message A{
    int32 ap = 1;
    int32 aq = 2;
  }

  // This field 'bz' refers to the *nested* message 'A' defined directly above it.
  A bz = 1;
  int32 by = 2;
}
```

The resulted zig file will error with ambiguous reference for field `B.bz`. The type of field `B.bz` should be `B.A`, but the current implementation write the type as `A` only. This PR change it so that the resulting type is `B.A`


Arwalk:
Thanks a lot for your contribution. I will look into it as soon as possible.

Arwalk:
Thanks a lot for your contribution, and sorry for the delay in reviewing this.

aaumar25:
Great. I am glad to help as well.

PR #103, opened 2025-08-01
Remove usingnamespace mixin

Arwalk:
strongly relates to #101 

pretty much the same thing overall but with full regeneration of generated files.

this will be then merged into zig-master.


src/protobuf.zig
```diff
@@ -1630,6 +1630,123 @@ fn stringify_struct_field(
     }
 }
 
+pub fn pb_json_parse(Self: type, allocator: Allocator, source: anytype, options: json.ParseOptions) !Self {
+    {
```
menduz, line 1634: is double { intentional?

menduz (approved):

src/protobuf.zig
Arwalk: nope !

Issue #96, opened 2025-03-24
Self referential types: identify cycles before encoding


Arwalk:
with the merge of #73 was identified that it could be nice to ensure that there are no cycles in the data to encode before trying to encode it.

Ideally, this check should only be performed when there _can_ be cycles, ensuring that we do not do comparisons for nothing on each encoding.

I've tried to do it on the branch "self-ref-check-cycle", but failed to make it work: my code seems to fail to compile and throws a segmentation fault on the opentelemetry structures.

If anyone is willing to take a look at this branch and see if they can find a solution, i'd love it. I'm not currently sure how it could work properly on all our structures but fail here.

PR #108, opened 2025-08-13
Cleanup & move towards writer pattern in wire encoding

mochalins:
This PR was motivated by attempting to patch the `zig-master` branch to support latest Zig master. Notably, the removal of managed `ArrayList` in `std` caused deep breakage; while it could be supported with enough band-aid fixes, I felt that it was time to rework `master` branch first to stop bitrot/pain in managing vastly different `master`/`zig-master` versions.

Refactors codebase to better follow Zig conventions and best-practices.

Goals:
- Explicit memory allocation -- move away from managed-container-esque storage of allocators to make all allocations unmanaged. Users should have to explicity pass allocators in signature for everything that can potentially (re/de)allocate.
- Remove usage of deprecated functions/types
- Follow best practices in struct initialization (e.g. default fields should not be set if changing them independent from each other can invalidate the struct).
- Remove all `ArrayList` output lines in bootstrap generator; we should be able to go directly from structured data in `GenerationContext` to a file with minimal allocations using writers.
- Zero-allocation wire encoding and minimal-allocation wire decoding making better use of writers/readers.
- Ensure bootstrap generator has zero memory leaks (it is currently leaking; not a big deal as it is a short-lived process, but still good to fix)
- Clearer allocation strategy (removal of all `Managed...` types)
- No `ArrayList` types in generated structures -- should use slices directly.

Progress towards #113, #111, and #110

### DIFF
--- a/bootstrapped-generator/main.zig
+++ b/bootstrapped-generator/main.zig
@@ -25,9 +25,8 @@ pub fn main() !void {
 
     try ctx.processRequest(allocator);
 
-    const stdout = &std.io.getStdOut();
-    const r = try ctx.res.encode(allocator);
-    _ = try stdout.write(r);
+    const stdout = std.io.getStdOut().writer();
+    try ctx.res.encode(stdout.any(), allocator);
 }
 
 const GenerationContext = struct {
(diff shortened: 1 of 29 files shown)


bootstrapped-generator/main.zig
```diff
@@ -730,41 +780,90 @@ const GenerationContext = struct {
...
+    var out: std.ArrayListUnmanaged(u8) = try .initCapacity(allocator, str.len);
+    defer out.deinit(allocator);
+    var writer = out.writer(allocator);
+    try writer.print("\"{}\"", .{std.zig.fmtEscapes(str)});
+    return out.toOwnedSlice(allocator);
+}
+
+test "self referential" {
+    _ = &GenerationContext.isMessageSelfReferential;
+    _ = &GenerationContext.analyzeMessageDependencies;
```
mochalins, line 868: Note @Arwalk: this is a dead function that was introduced in https://github.com/Arwalk/zig-protobuf/commit/37591ae7e0dae5bb6b288059c14c4ae7ad2da7ea with no callers. Unsure of what your intentions were with this function.

Arwalk: I'll be honest and have no clue about it. If it's dead, you can remove it.

mochalins: I believe it may be related to #96, although I'm not sure; I'll keep it in for now until the final merge review, and we can determine whether it should stay then.

mochalins:
Btw @Arwalk; if you accept/merge PRs into `feature/ziggify`, could you do a rebase rather than the default merge/squash please? That'll make it much easier for me to keep working on dependent PRs.

Issue #114, opened 2025-08-14
Better clarify allocation strategies, removing all `Managed...` types


mochalins:
Use of `Managed...` types seems similar to C++ RAII-isms, marking memory as owned/borrowed/static to make bulk deinitialization at end-of-scope simpler to manage.

Instead, we should rework allocation strategies from first-principles to pass more explicit control to the user; the user should be able to easily track which resources require freeing, with no need for any `Managed` type overhead. This likely involves reducing nested/inner allocations whereever possible, potentially making better use of arenas, and increasing reliance on writer/reader patterns rather than result-holding slices.

Arwalk:
Can i ask you to look if the `init` function will still be necessary after your work? I've been told that it could now be completely unnecessary, considering there could be a default value every time, making `.{}` the only init we need.

mochalins:
I hear your concerns, and I think there might be a misconception.

Using slices directly does not mean you now have to copy every slice passed in; you can "move" strings in just like before, as long as it was allocated with the same allocator used for the message. This works in the other direction as well -- after decoding a message, you can "move" strings out by simply using the string elsewhere and setting the string inside the message to an empty `&.{}`.

This is the key point; we're moving control of memory from the library to the user. This general principle is pretty fundamental to Zig's zen, and is the approach taken by essentially all (well designed) third party libraries and the standard library.

If we continue to use `ManagedString`/`ManagedStruct` devices, the overhead is quite insane; every single string will require potential additional pointer dereferencing, and we will be storing an additional tag + allocator interface's worth of memory with _every single string_. Not to mention the additional branching required on every single string access.

If we continue to use managed types, this overhead is _impossible to work around_; it's a penalty all users are forced to endure.

However, if we use raw slices, then users may still be able to replicate previous behavior with even better performance than what the library would ever be able to do with managed types; they can implement a custom deallocation strategy that looks up a table of static slices/slices from mixed allocations, either implementing the full deinit for max performance or simply setting those slices to empty before calling the message's deinit.

> Also getting rid of const slices have a huge penalty on my protobuf-based profiler

I'm a little suspicious about this one; in case you weren't aware, our previous encoding strategy already required essentially reallocating every single string (const, copied, or moved) every single time you sent a message, as it wrote to an output slice before then passing that slice to a writer.

Just with the direct-writer changes alone we should now be doing _strictly less_ copying _even if_ you decide to now duplicate every single string.

Similar changes are upcoming to decode.

[...]

PR #115, opened 2025-08-14
Remove all `Managed` types

mochalins:
Progress towards #114 , depends on #108


src/protobuf.zig
```diff
@@ -665,16 +550,86 @@ fn dupe_field(original: anytype, comptime field_name: []const u8, comptime ftype
...
+                .pointer => |p| {
+                    comptime std.debug.assert(p.size == .one);
+                    const result = try allocator.create(p.child);
+                    result.* = try @field(original, field_name).dupe(allocator);
+                    return result;
+                },
+                // Handle submessage in "oneof"; non-optional struct in union.
+                // Submessages in unions may be non-optional in proto3 and
+                // editions, as submessages always have an explicit presence.
+                // As such, an "empty" submessage cannot be sent, and the lack
```
Arwalk, line 623: `may be non-optional in proto3 and 2 editions` i guess? Missing the 2. 

Side note: do not bother too much with proto2 compliance, we're aiming for proto3 before all.

src/protobuf.zig
```diff
@@ -1747,42 +1775,49 @@ pub fn pb_jsonStringify(Self: type, self: *const Self, jws: anytype) !void {
 
 pub fn MessageMixins(comptime Self: type) type {
     return struct {
```
Arwalk, line 1777: Let's stop supporting this useless piece of code and remove it !

Arwalk (approved):

src/protobuf.zig
mochalins: Hmm, I should improve the wording of this comment.

As an aside, the "editions" here means protobuf editions (2023+), and not proto2; I'm also ignoring proto2 beyond the minimum compatibility required for passing our test suite.

PR #116, opened 2025-08-14
Reorganize functions/namespaces, optimize message duplication and wire tag encoding

### LINKED ISSUE #87: Move all JSON code to a different file
JSON code consumes a good percentage of the total lines, we may move it to its own file for clarity.

mochalins:
Some reorganization to prepare for decoding changes; made a few small optimizations while I was reorganizing. 

Closes #87.


src/wire.zig
```diff
@@ -0,0 +1,206 @@
...
+    /// fixed64, sfixed64, double
+    i64 = 1,
+    /// string, bytes, embedded messages, packed repeated fields
+    len = 2,
+    /// group start (deprecated)
+    sgroup = 3,
+    /// group end (deprecated)
+    egroup = 4,
+    /// fixed32, sfixed32, float
+    i32 = 5,
```
Arwalk, line 19: Can we rename `i64` and `i32` as `fixed64` and `fixed32` respectively ? I'm not fond of the collision with the type `i64` / `i32`

Arwalk: (i know the protobuf documentation calls them like this but it should be explicit enough anyway)

src/wire.zig
```diff
@@ -0,0 +1,206 @@
...
+};
+
+/// Record tag.
+pub const Tag = packed struct(u32) {
+    /// Wire type.
+    type: Type,
```
Arwalk, line 25: rename to `wireType` avoid collision with keyword `type`

src/wire.zig
```diff
@@ -0,0 +1,206 @@
...
+    /// Field number.
+    field: u29,
+
+    /// Encode tag to byte stream. Returns number of bytes used for encoding,
+    /// which is guaranteed to be between 1-3 inclusive.
```
Arwalk, line 30: You mean 1 and 5 inclusive?

src/wire.zig
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        const invalid_wire_type = (raw_result & 0x7) > 5;
+        if (invalid_wire_type) {
+            @branchHint(.cold);
+            return error.InvalidTag;
+        }
+
+        return @bitCast(@as(u32, @intCast(raw_result)));
+    }
```
Arwalk, line 103: Can't we just use a decode varint that returns a u32 and make a Tag out of it? Is there any benefit in duplicating decoding logic logic here?

src/protobuf.zig
```diff
@@ -259,6 +340,7 @@ fn writePackedStringList(
     comptime field: FieldDescriptor,
     value_list: anytype,
 ) std.io.AnyWriter!void {
+    // TODO: find examples about how to encode and decode packed strings. the documentation is ambiguous
```
Arwalk, line 343: Create an issue for this please, thanks for pointing it out.

Arwalk (changes requested):
Some slight changes/questions, but overrall going in a good direction.

Unrelated to your change: i found a log line 894 in protobuf.zig `std.log.err("list of strings deinit", .{});`, could you take this opportunity to remove it?

src/wire.zig
mochalins: While I can rename to `fixed32` and `fixed64`, can you clarify what you mean by collision? As these are fields, they are guaranteed to always be safely disambiguated; in all usage instances, they must be preceded by `.`, so even naive regex syntax highlighting won't confuse these field names with the primitives (with the singular exception of their declaration location, as you can see with the GitHub highlighter here).

src/wire.zig
mochalins: Actually, no, there isn't a benefit in duplicated decode logic! This separate logic stemmed originally from when I misunderstood the protobuf spec and assumed the max number of fields would be equivalent to the max field number. As such, it seemed reasonable to create a custom wire decode as the sole scalar type that must fit in a `u16` equivalent. 

However, after catching this incorrect assumption, it is certainly better to reuse scalar decoding. As this PR doesn't yet have the new `decodeScalar`, I'll make this change in #119

src/wire.zig
Arwalk: The problem is not about the tooling in itself, even if you pointed out yourself that even some tools fail to highlight the thing properly like in github.

The problem is readability of the code as a human reader. You spent a significant amount of time on this code, so you currently have a good mental model of what is happening in it. But someone that reads the code for the first time or only occasionally will have a "double take" moment. The reader will stumble upon this `.i64` in some part of the code, will be interrupted by seeing something that's not usually associated, and will have to stop his current analysis and thought process just to understand why this `i64` integer type is suddenly preceded by a dot.

This kind of double-take/back and forth make the code less readable in the long term, it's overall a good practice to not name something with a keyword that is used in an entirely other context. 

Sorry for being nitpicky about this.

side note: we can add a little comment over `fixedSize64` that says something along the line if `// i64 on protobuf spec`

side side note: as your code is closer to some part of the specification, it could be cool to add a comment with a link to the part of the specification you're referring to.

src/wire.zig
Arwalk: agree with naming

Arwalk: agree with naming

Arwalk (approved):

PR #119, opened 2025-08-17
Rework decoding

mochalins:
Requires #116

### DIFF
--- a/src/wire.zig
+++ b/src/wire.zig
@@ -482,8 +482,17 @@ pub fn decodeRepeated(
             try result.append(allocator, try .init(allocator));
             errdefer result.items[result.items.len - 1].deinit(allocator);
             const msg = &result.items[result.items.len - 1];
-            try decodeMessage(msg, allocator, reader, options.bytes.?);
-            return options.bytes.?;
+            const consumed = try decodeMessage(
+                msg,
+                allocator,
+                reader,
+                .{ .bytes = options.bytes },
+            );
+            if (consumed > options.bytes.?) {
+                @branchHint(.cold);
+                return error.InvalidInput;
+            }
+            return consumed;
         },
     }
 }
@@ -508,7 +517,11 @@ test decodeRepeated {
             },
         );
         try std.testing.expectEqual(bytes.len, consumed);
-        try std.testing.expectEqualSlices(u32, &.{ 3, 270, 86942 }, list.items);
+        try std.testing.expectEqualSlices(
+            u32,
+            &.{ 3, 270, 86942 },
+            list.items,
+        );
     }
 }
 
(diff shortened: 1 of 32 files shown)


mochalins:
Rebased on 32-bit tag type fix in #116

src/wire.zig
```diff
@@ -151,55 +159,967 @@ pub const ZigZag = struct {
...
+                        const oo_field_ti = @typeInfo(oo_field.type);
+                        switch (comptime inner_desc.ftype) {
+                            .scalar => |scalar| {
+                                if (comptime scalar.isSlice()) {
+                                    std.debug.assert(tag.wire_type == .len);
+
+                                    // `oneof` fields are always non-optional
+                                    // as `oneof` has explicit presence.
+                                    std.debug.assert(
+                                        oo_field.type == []const u8,
```
mochalins, line 869: @Arwalk Apologies for not noticing this sooner, but while rebasing I realized that the standard library already has an example of using field names that conflict with keywords; it seems Zig prefers using the maximally clear/precise name even in case of collisions, such as `type`. 

Would this change your thoughts on the `.i32`/`.i64`/`.type` feedback in #116 at all?

Arwalk: >Would this change your thoughts on the .i32/.i64/.type feedback in https://github.com/Arwalk/zig-protobuf/pull/116 at all?

Not really, because in the zig library, when type is used, it still refers to a value representing a `type` as in the typesystem of zig. This might be the only exception i'd allow, where the "type" field is actually what a "type" is in the whole context of zig.

Sorry, i'm a stubborn person :)

mochalins: > in the zig library, when type is used, it still refers to a value representing a type as in the typesystem of zig

@Arwalk Hmm I don't actually think that's true; see, for example, 

https://github.com/ziglang/zig/blob/1a5cf072a8894976c1cd16d7132af671c7cdbb4d/lib/std/os/uefi/protocol/device_path.zig#L12

There's actually more cases of this, both with the `type` keyword but also with the `align` keyword, and others.

Haha I know I mentioned I didn't find the naming too important, but I'll have to recant that statement. While looking into this further, I've developed a stronger preference; I think it's better to use the names explicitly specified in `protobuf` spec rather than creating custom names because of a Zig keyword's existence.

More examples in the newest `aarch64` backend's `encoding.zig`, ELF linker, even in [resinator's CLI](https://github.com/ziglang/zig/blob/1a5cf072a8894976c1cd16d7132af671c7cdbb4d/lib/compiler/resinator/cli.zig#L97).

LMK what you think, perhaps this can be opened in a separate issue discussion?

Arwalk: I tend to agree with using the domain-specific language as much as possible, but the clash with the domain and the language keywords are really a problem for me. I honestly think that the examples you provided here are simply mistakes (especially everything related to the word "type", that is such a generic word that it easily leads to problems imho).

I think that is also because you explicitely want to expose the internals of the protobuf library so that you can do your "stream parsing". My opinion on this would that we should have a dedicated interface for this, with its own keywords and use. The point of the library is to abstract protocol buffers itself imho. I said it earlier that i'd like us to think about an interface for the kind of approach you're doing on your side.

Create an issue if you want, but i'm not really enclined to debate further on this. Sorry.

Arwalk: I can't force you to contribute. Once again, i'm sorry for my inability to review/maintain in a timely manner.

Issue #112, opened 2025-08-14
Use writers directly in bootstrapped generator, removing all output lines storage


mochalins:
Bootstrapped generator currently uses `ArrayList`s to store all output lines. Only after all lines are generated does it then write to file/`stdout`. 

This generally causes worse performance (unnecessary allocations/copies/cache hits/resizes), difficult-to-resolve memory leaks (e.g. some lines are embedded constant strings that cannot be freed, other lines are dynamically allocated; truly making sure there are no leaks would involve either using an arena allocator which would be potentially catastrophic for memory use during resizes, or tracking ownership of every line embedded in the `ArrayList` à la `ManagedString`, which adds significant overhead), and is less flexible to use than writers.

All `ArrayList` output lines should be removed so that the generator can instead write directly to writers.

Issue #111, opened 2025-08-14
Ensure up-to-date adherence to general Zig conventions


mochalins:
This includes naming, line lengths/formatting, removing use of deprecated functions/definitions, and common patterns (e.g. struct initializations/default fields, using RLS for decl access, etc.)

PR #120, opened 2025-08-20
Remove `init` function from generated messages 

mochalins:
Depends on #119. 

`init` function is removed from generated messages, instead preferring to use Zig's standard struct initialization.

While `proto2` compatibility is not a goal, this also better enforces `required` fields; users are now forced to provide required fields with struct initialization rather than silently setting them to the protobuf field type default.

Previous:
```zig
var my_message = try Foo.init(allocator);
```

Now:
```zig
var my_message: Foo = .{};
```


mochalins:
@Arwalk NB: Zig 0.15.0 release was just recently tagged; while not all of the original goals were perfectly accomplished (notably #112, although some minor work is left in #111 and some potential future enhancements left in #114), I feel that this PR reaches a state where I'd be comfortable calling `feature/ziggify` "finished", and moving on to porting for 0.15.0. Lmk what you think!

menduz:
I wouldn't remove the default values. Non trivial protobufs will take so much labor to write packages to fill-in default values, and changes to .protos will likely break the compilation.

hawkbee:
come across [this discussion](https://ziggit.dev/t/why-this-array-syntax/11538/37), the [Decl Literals](https://ziglang.org/download/0.14.0/release-notes.html#Decl-Literals)  maybe the prefer way to initialize a struct.

Arwalk (approved):

PR #126, opened 2025-09-17
Create CONTRIBUTING.md, add Pull Request Etiquette

menduz:
Taken from https://gist.github.com/mikepea/863f63d6e37281e329f8, links to https://docs.github.com/en/site-policy/github-terms/github-community-code-of-conduct

This guide ranks very high in the list of most positively impactful things I've seen in engineering teams 

It describes why and how to contribute using pull requests, sets a slim set of requirements to 1) encourage quality and 2) help making reviews easier and faster

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,7 @@ Remember that your commit message must survive the ravages of time. Try to link 
 
 ** Each commit message should include the reason why this commit was made. Usually by adding a sentence completing the form 'So that we...' will give an amazing amount of context to the history that the code change itself cannot **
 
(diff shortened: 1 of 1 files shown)


Arwalk:
Should we also refer to [conventional comments](https://conventionalcomments.org/) ? It sometimes help smoothing discussions between perfect strangers over reviews. I often forget to use them myself, but i believe it's not a bad thing.

Arwalk:
Also, should we adhere to conventional  commits  https://www.conventionalcommits.org/en/v1.0.0/ ?

Arwalk:
> I would enforce conventional commits, and suggest using conventional comments. Commits are the written story of the repository and the conventional commits are a good way to condense what's in the commit, also widely adopted.

Im agreeing on this.


> For collaboration, I'd leave them as a suggestion. I tend to do `nit: ` or `style: ` prefixed messages in a professional environment. I'll leave you the final decision

Fine with that too. When we're working together we don't have to force ourselves to do it, as we start to know each other, but we should try to adhere to them when working with less usual contributors.

menduz:
changes added @Arwalk

PR #138, opened 2025-10-05
use build.zig.zon to download protoc

menduz:
Remove dependencies on `curl` and `unzip` to leverage zig's lazyDependencies.

This approach also ensures that the downloaded protoc files are not altered via hash verifications.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,11 @@ debug
 .specstory/*
 test.data
 benchmark/generated/**/*
+
+# nix devenv
+devenv.nix
+devenv.yaml
+devenv.lock
+.envrc
+*.nix
+.devenv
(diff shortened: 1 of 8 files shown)


menduz:
> Does this still allows for externally-provided protoc? nixos might require it, as far as i remember.

No, that option is no longer allowed. But this approach works on nixos, I've developed this PR in nixos. Moreover, I was having trouble related to cwd and PATH variables: unzip & curl commands not found in nix. It was working for more relaxed distros.

Arwalk (approved):

Arwalk:
>  I think 3.0.0 never really worked because of this path issues, we may not need a 4.0.0

I had toy projects to test the dependency management with zbench to properly get protoc and work out of the box on 3.0.0, so I think it's a specific case for some linux systems. Gonna make a release anyway.

PR #146, opened 2025-11-29
build: build protobuf locally instead of downloading binary blobs

### LINKED ISSUE #144: Please disclose that zig-protobuf downloads and runs binary blobs from Google
Nothing in the README indicates that this Zig module downloads `protoc` binaries from Google and then runs them, which is a serious security and trust issue. It would be nice to give some kind of warning, as "a protobuf 3 implementation for zig" implies that there wouldn't be unverifiable binary blobs of C++ downloaded from Google, and in fact, you [aren't even allowed](https://github.com/Arwalk/zig-protobuf/pull/138#issuecomment-3369261238) to use anything else. Please do not download and run binary blobs on my computer without my consent. Thank you.

Jan200101:
fixes #144 and potentially makes this project available to more targets
only tested on Fedora 43 x86_64 by running both `zig build -Dtarget=x86_64-linux-gnu test` and `zig build -Dtarget=x86_64-linux-gnu bootstrap`
(target is used to make Zig not use any system resources but not needed)

- add local package to compile abseil (dependency of protobuf)
- add local package to compile protoc, libprotoc and libprotobuf
- rework build infrastructure to pass targets and optimization down, prevent any configure time path dependency and proper graph generation.
- also turn `RunProtocStep.create` into a wrapper that calls `RunProtocStep.createWithGenerator`, no need to duplicate things.

### DIFF
--- a/pkg/abseil/build.zig
+++ b/pkg/abseil/build.zig
@@ -19,6 +19,8 @@ pub fn build(b: *std.Build) void {
 
     if (target.result.os.tag == .windows) {
         abseil.linkSystemLibrary("dbghelp");
+    } else if (target.result.os.tag == .macos) {
+        abseil.linkFramework("CoreFoundation");
     }
 
     abseil.addIncludePath(upstream.path(""));


Jan200101:
if needed I can also adapt this PR to check for system capabilities i.e. if `protoc` is installed on the system to avoid long compilation times

Arwalk:
Hey there @Jan200101 , thanks for this PR. I really like the idea of providing protoc through compilation with zig's toolchain.

Though i'm starting to believe that this could be very beneficial for the zig community to provide this functionnality through an external package.

I'm thinking that we could create a zig-protoc package , maybe under the [zig-protobuf](https://github.com/zig-protobuf) organization (that is currently unused, i wish i had the idea that my toy project would get some interest), that provides a few utilities:

- a `downloadProtocBinary` function that does what is currently done on the project, using lazy dependencies to download and extract protoc, and returns a build step that can be used to find this downloaded protoc.
- a `buildProtocFromSource` function that does what you are proposing, and returns a build step that can be used to find this built protoc.
- a `findSystemProtoc` function that checks if there's already a protoc available on the system, and returns its path if available.

Our own `RunProtoc` step should only need a path to protoc to run with the arguments provided. If we could make it so that RunProtoc uses that path, this would allow the user all the flexibility they would need, as in:
- Using a downloaded protoc for easy setup
- building from source if preferred
- Using an already installed protoc
- Or even giving a path at build time directly if the use want to alternate between protoc versions for some reasons.

`zig-protobuf` would depend on this external package, and exposes its utilities for users.

I think having this little package outside of this project would be very cool. Especially if someone wants to make another protobuf implementation in zig.

What do you think of this idea? @menduz any opinion ?

Would you like to participate in this project @Jan200101 ?

Jan200101:
I have no interest in maintaining this project long-term and don't have any personal projects that use protobuf in Zig.

https://github.com/AllYourCodebase/ exists as a community org for porting C/C++ projects to the zig build systems so that zig projects can easily consume them, might be better to move the protobuf and abseil impls there.

Arwalk:
> I have no interest in maintaining this project long-term and don't have any personal projects that use protobuf in Zig.

I absolutely understand. Even if you don't want to maintain, I'd like to take some inspirations (aka copy-paste) some of your work here for this project, but I would like you to keep the credits for the time and effort (aka work) that you have provided on this PR.

I really don't think i'll merge this PR, as i'd like to do this in a separate project as I said, but i wish to keep the trace of your contribution.

> https://github.com/AllYourCodebase/ exists as a community org for porting C/C++ projects to the zig build systems so that zig projects can easily consume them, might be better to move the protobuf and abseil impls there.

Cool project. I think i'll start on my side, then see if they are interested in moving it in their project. Thank you.

Arwalk:
I feel like the discussion became a bit heated, but i have absolutely no idea why. We have responded positively to the request, even proposing to set it apart from this repo to empower other projects to use protoc easily, and overall have been really understanding of the needs and reasons around not wanting to download protoc binaries right away. I feel like there were some friction about "which choice is better", but the whole point of what was proposed in the end was about giving full power to the library user.

@Jan200101 i'm sorry if somehow someone, me included, made you feel uncomfortable with this PR. Your contribution here was truly valuable. If you are willing to port some of it to https://github.com/zig-protobuf/zig-protoc, it would make me very happy, but don't feel pressured. I have no right to ask anything from you. Have a nice day, and thank you again.

Jan200101:
don't worry, in no way did the responses pressure or discourage me or make me feel like I needed to close this PR, I did so because I wanted to focus on moving the code to compile protobuf into its own project which can then be more easily consumed by other projects, including this one.
Once I verify that everything works as intended I'll open a new PR that downloads by default and otherwise compiles it with an option to always build.

Arwalk:
> I did so because I wanted to focus on moving the code to compile protobuf into its own project which can then be more easily consumed by other projects, including this one.

Great, i'm conting on you for letting us now when you're done. Thank a lot again.

PR #149, opened 2025-12-09
Service and RPC generation

inge4pres:
### Reason for this PR

Addresses #39 

### Details

RPC methods are generated as function pointers and then the struct methods are dispatched via the generic types that servers will implement.

`std.Io` Reader and Writer are already in the function signature, opinionated design choice so let me know if you agree.

### DIFF
--- a/src/protobuf.zig
+++ b/src/protobuf.zig
@@ -718,7 +718,7 @@ fn dupeField(
             }
             return list;
         },
-        .packed_repeated => |_| {
+        .packed_repeated => {
             const capacity = @field(original, field_name).items.len;
             var list = try @TypeOf(@field(original, field_name)).initCapacity(allocator, capacity);
 


Arwalk (commented):
I made a quick pass, these are some really good starting points. Do you think you could provide some examples/tests about how the readers and writers are used in the stream interfaces ? Of course even very naive implementations are enough.

I'll try to dive into this whenever I can.

tests/generated/tests/service.pb.zig
```diff
@@ -0,0 +1,355 @@
...
+
+/// Service demonstrating all streaming patterns
+pub fn StreamingServiceImplementations(comptime ServerContext: type) type {
+    return struct {
+        /// Server streaming RPC - client sends one request, server sends stream of responses
+        ServerStream: *const fn (context: *ServerContext, request: StreamRequest, writer: *std.Io.Writer) anyerror!void,
+        /// Client streaming RPC - client sends stream of requests, server sends one response
+        ClientStream: *const fn (context: *ServerContext, reader: *std.Io.Reader) anyerror!StreamResponse,
+        /// Bidirectional streaming RPC - both client and server send streams
+        BidiStream: *const fn (context: *ServerContext, reader: *std.Io.Reader, writer: *std.Io.Writer) anyerror!void,
```
menduz: I think the correct abstraction for RPCs with protobuf lies in the neighbourhood of messages and not writers:

```suggestion
        /// Server streaming RPC - client sends one request, server sends stream of responses
        ServerStream: *const fn (context: *ServerContext, request: StreamRequest, writer: WriterDelegate(StreamResponse)) anyerror!void,
        /// Client streaming RPC - client sends stream of requests, server sends one response
        ClientStream: *const fn (context: *ServerContext, reader: AsyncIterator(StreamRequest)) anyerror!StreamResponse,
        /// Bidirectional streaming RPC - both client and server send streams
        BidiStream: *const fn (context: *ServerContext, reader: AsyncIterator(StreamRequest), writer: WriterDelegate(StreamResponse)) anyerror!void,
```

And ServerContext, at a higher level, should be in charge of adapting the messages to the Io interfaces. Otherwise, every method of the entire RPC implementation would need to be aware and implement the adapter for HTTP/2, headers and trailers (or whatever transport you choose).

The suggested way also removes the need of thinking about Io at all during tests, because you can test the RPC using pointers to the messages, leaving the transport and the serialization completely out of the center stage.

menduz: https://github.com/ziglang/zig/blob/738d2be9d6b6ef3ff3559130c05159ef53336224/lib/std/Io.zig#L1487 queue is the sound abstraction for thread safe streams in zig

inge4pres: oh btw, queues are not yet released, I am using 0.15.2 and can't see them in Io.
are we targeting a specific dev version of 0.16 or latest main works?

inge4pres: > I'd say it makes sense to move the branch to the stablest version supporting Queue, even if it is untagged

IIRC there's a `zig-master` branch that targets the unreleased version of Zig.
The point is that keeping `master` on a Zig version that is generally available allows pinning to `master` all projects that are on a released tag.
I am biased here, because I have several projects on 0.15.2, but thinking in more broader terms, if we merge to `master` a PR that uses unreleased Zig features, aren't we breaking a promise of compatibility?

menduz: you are right, if Queue is not at 15.2 then we should merge the pr to zig-master

Arwalk:
where are we on this one @inge4pres ? Want to move to zig-master? You can also leave my comments on anyerror, i'm sure i'm being picky for nothing.

I'm not trying to push you on working on this of course, it's whenever you have the time.

Arwalk:
> > where are we on this one @inge4pres ? Want to move to zig-master?
> 
> hey @Arwalk thanks for the ping and apologies for the delay, I'v been busy and could not get back to this PR.
> 
> I'll move the target branch to `zig-master` and take care of the comments where possible.
> Given the upcoming holiday season i don't expect reviews or quick follow ups but I am surely committed to making the change land 👍🏼 

Don't stress yourself. Focus on the holidays.

tests/test_services.zig
```diff
@@ -0,0 +1,234 @@
...
+const TestServerContext = struct {
+    allocator: std.mem.Allocator,
+    call_count: usize = 0,
+
+    // Error types required by the service implementations
+    pub const UnaryCallError = error{TestFailure};
+    pub const AnotherCallError = error{TestFailure};
+    pub const ServerStreamError = error{TestFailure};
+    pub const ClientStreamError = error{TestFailure};
+    pub const BidiStreamError = error{TestFailure};
```
menduz: Seeing this errors defined at this level of abstraction made me notice that the current layers are leaking many details upwards the RPC protocol adapters.

I don't see the point of having this errors defined at this level of abstraction. Because: who catches them? The caller of the handler functions will _always_ be the implementation of the RPC protocol. I doubt these are exhaustive or known for each RPC option. 

Each RPC implementations will have a hard time compiling with user-defined errors, it will simply not know what to do with them or what they are. 

Maybe the `service.SimpleServiceImplementations` should be an interface to be used by wrappers and it should receive the error enums. Also, the `context` field seems self referential, the layers are again leaking by having structs pointing to parent structs.

Here is a draft that I think may solve the issue I'm describing: the generated function `SimpleServiceImplementations` is is dropped, and `SimpleService` now receives a "UserDataType" and "ErrorEnums", and returns a VTable struct.

```diff
-pub fn TestService(comptime ServerContext: type) type {
+pub fn TestService(comptime UserDataType: type, comptime ErrorEnums: error) type {
    return struct {
+       // TODO: include service metadata to enable clients generation

        Foo: *const fn (userdata: *UserDataType, request: FooRequest) ErrorEnums!FooResponse,
        Bar: *const fn (userdata: *UserDataType, request: BarRequest) ErrorEnums!BarResponse,
    };
}
```

```zig

// http-based rpc

const HttpErrors = error { InternalServerError };

pub fn HttpServer(comptime ServiceGenerator: anytype, comptime userDataType: Type) anytype {
  // @assert(ServiceGenerator is a function)
  return struct {
    allocator: *std.mem.Allocator,
    userdata: *UserDataType,
    vtable: *VTable,

    fn listen(self: *@Self(), io: std.Io) {
       // wire this implementation
    }

    pub fn init(allocator, userData: *UserDataType, implementation: *VTable) @This() {
      // here implement the initialization code for this rpc adapter
    }

    pub const VTable = ServiceGenerator(UserDataType, HttpErrors);
  }
}


// to be implemented as

struct UserData {
  .call_count: i32
};

const MyServer = HttpServer(generated.SimpleService, UserData);

var userData: UserData = .{.call_count = 0};

const httpHandlerImpl: MyServer = .init(
    allocator, 
    &userData, 
    .{ // vtable
       .UnaryCall = &unaryCallImpl,
       .AnotherCall = &anotherCallImpl,
    },
)

// actual handlers
fn unaryCallImpl(userData: *UserDataType, request: service.UnaryRequest) HttpErrors!service.UnaryResponse {
  // ...
}

// TODO: fn anotherCallImpl ommitted

// TODO: httpHandlerImpl.bind(io)
```


Note that all this code is only enough to generate servers, clients would require a different approach, maybe involving a lot of comptime. It may be good to leave clients for a new PR

inge4pres: > if your handler has a error.MyCustomError, the grpc server will definitely not know what to do with that and wont compile unless we cast the function to anyerror and hope for the best.

Apologies if I was vague in my previous comment, I'm suggesting that the error is provided by us in the library, a single error type for all services that is assigned to all signatures automatically.
It's simpler for the library and also for the users.
I had initially proposed it as an alternative to `anyerror` but @Arwalk counter-proposed user-provided errors (https://github.com/Arwalk/zig-protobuf/pull/149#discussion_r2617577323).

As you mentioned, they come with a little overhead for users to implement, so should we revisit a library-provided error set?

I am thinking to something like this is protobuf.zig

```zig
const ServiceError = error {
  RPCError,
}
```

and then in bootstrapped-generator/main.zig we only include `protobuf.ServiceError.RPCError` in the generated methods' signature.

We would keep the `ServerContext` as a mean to deliver the business logic of the server, but we would simply allow the service to return a generic error, leaving the underlying implementation dealing with it.
This would also remove the `requireDecls` helper needed in the current code.

> For the sake of shipping, if you feed the comments of the pr to a coding agent I assure there is enough context to let them code the changeset for us.

I value more if we take a small step towards progress than a full-fledged solution, because we apparently don't have a clear path ahead, we don't know if users of the library find it ergonomic.
As a user of the library myself I am not yet clear on what's the best path when implementing services.
Plus, this PR is to a dev release of Zig and I don't expect many users to be running rolling releases on their codebases.
The way I see it speed is not a goal, while it is a goal to get something simple done, in order to address feedback in the future.

menduz: I'm re reading all the comments, I missed the error ones, sorry for that, we are having the same conversation in two different places. @Arwalk I allowing user-defined types is a no-go because the RPC implementations will never know what to do with errors. And the RPCs are the callers of this functions (all this is stated above with more detail).

I think settling this argument with a single library-provided error will be enough to start. We should be clear with users that the interface is not stable as the interface of the messages, and its final shape will depend on how the first RPC implementations turn out to be. We can have multiple comptime-driven generators for this kind of interfaces anyways. As long as we declare a couple of structs and consts like we do for messages (pending, may be other PR).

@inge4pres Thanks for linking to the comment, I completely missed it. We agree on a library-provided error to start and ship this. I still think that having two structs for context-proxying can be reduced to one single comptime function returning a struct of `fn(userdata: *UserDataType, ...)` arguments. I also lack the language to describe how it will be beneficial, probably a concrete example will illustrate better than I can explain. I'll fork your branch to do it

menduz: Here it is, https://github.com/Arwalk/zig-protobuf/commit/dc9c021ad2c216f6fc5867f00a06ae6ef30de752 I think it has the best of both worlds. User (or library) provided error set and one fewer container struct for context, uses the same "userData: *UserData" pattern that is repeated all over the zig std library

Arwalk: You guys are doing great. Thank you for all your work. Sorry for being very absent in general.

menduz: I will be exclusively on mobile until friday. Can't fin the option to pick a branch from a fork from the app... I don't mind the authorship of the commit if you can apply the patch locally and push to your branch. 

Some fixes for std.io for zig 16.0-1600 were merged to zig-master already. My branch also fixes a new signature change for fs.path.relative

menduz: I'm having a hard time creating a PR to your branch, would you cherry pick instead?

Thanks

inge4pres: Yep going for it and incorporating the latest Zig dev build

inge4pres:
Updating the README wrt to the latest changes for the service generation

menduz:
Thanks to you!

I think it should be a separate project for clarity, we will definitely backlink. Also I bet the Zig team will love to see how gRPC is born out of comptime code, let us know if you need assistance. I have never dug into the grpc implementation, I only know it from a 10000ft view, but maybe can make myself useful.

PR #150, opened 2025-12-22
std.meta.intToEnum -> std.enums.fromInt

agagniere:
`std.meta.intToEnum` was deprecated in 0.15 and is removed for 0.16

This patch fixes `zig build` but `zig build test` fails because of too much eval branches, suggestions welcome

### DIFF
--- a/src/wire.zig
+++ b/src/wire.zig
@@ -529,7 +529,7 @@ pub fn decodeMessage(
         };
         consumed += tag_c;
 
-        @setEvalBranchQuota(40_000);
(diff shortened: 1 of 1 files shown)


Arwalk:
@agagniere & @menduz , the eval branch change has been "fixed" in https://github.com/Arwalk/zig-protobuf/commit/1c7ca085dffaa5fa42e4e8d3872d22dc509d2eb3 , but i'm creating an issue on how to handle this in a more graceful way.

PR #161, opened 2026-02-16
fix(build): allow for relative build paths

mehcode:
- Set `minimum_zig_version` to allow ZLS to work automatically in VS Code
 - Flatten `dirExists` and `fileExists` into `pathExists`
 - Allow for relative build paths

The goal here is to support generating code into `.zig-cache/` and importing as an anonymous module. The following snippet works in my project after this change:

```zig
    // Obtain a handle to a generated directory in .zig-cache where we can write the output of `protoc`.
    const wf = b.addWriteFiles();

    const protoc = protobuf.RunProtocStep.create(protobuf_dep.builder, target, .{
        .destination_directory = wf.getDirectory(),
        .include_directories = &.{b.path("proto")},
        .source_files = &.{
            b.path("proto/One.proto"),
            b.path("proto/Two.proto"),
        },
    });

    // After `protoc` runs, there should be a single file named `PACKAGE.pb.zig` in the output directory.
    const root = try wf.getDirectory().join(b.allocator, "PACKAGE.pb.zig");

    // And make it available as an anonymous import to the root module,
    // so that it can be imported with `@import("PACKAGE.pb")` from any source file.
    mod.addAnonymousImport("PACKAGE.pb", .{
        .root_source_file = root,
        .target = target,
        .optimize = optimize,
    });

    protoc.step.dependOn(&wf.step);
    compile.step.dependOn(&protoc.step);
```

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -4,7 +4,7 @@
 
     .fingerprint = 0x99fac5be6a36efd1,
 
-    .minimum_zig_version = "0.15.1",
+    .minimum_zig_version = "0.16.0-dev.2510+bcb5218a2",
 
     .dependencies = .{
         .@"protoc-linux-64" = .{
--- a/utils/build_util.zig
+++ b/utils/build_util.zig
@@ -6,15 +6,8 @@ const Io = std.Io;
 pub const PROTOC_VERSION = "32.1";
 
 // File system utilities
-pub fn dirExists(io: Io, path: []const u8) bool {
-    var dir = Io.Dir.openDirAbsolute(io, path, .{}) catch return false;
-    dir.close(io);
-    return true;
-}
-
-pub fn fileExists(io: Io, path: []const u8) bool {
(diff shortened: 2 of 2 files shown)


Arwalk:
Hello @mehcode , thanks a lot for this PR.

Do you know if this is possible for 0.15.2?

mehcode:
Yeah it's doable. I can send a patch to mirror this on `master`.

Arwalk:
> Yeah it's doable. I can send a patch to mirror this on `master`.

Thank you. I prefer that new developments are made on master, then ported on zig-master, whenever possible.

Thanks a lot for your patience and interest!

Arwalk:
Hey there @mehcode , with zig 0.16.0 we're moving the master branch to 0.16.0

Do you think you could rebase and change the branch target to master? This should solve everything and could be merged fast.

Arwalk:
@mehcode are you ok with me grabbing some of your code to port it myself to master? I understand that you might not have the time nor motivation to take care of this, no problem with that at all.

PR #163, opened 2026-02-21
Fix json oneof shadowing

### LINKED ISSUE #147: Incorrect JSON serialization for otel protos
Hello! big fan of the library, however, I'm running into an issue with json serialization.

```zig
        const resource_attributes = [_]KeyValue{
            .{ .key = "service.name", .value = .{ .value = .{ .string_value = self.service.name } } },
            .{ .key = "service.instance.id", .value = .{ .value = .{ .string_value = self.service.instance_id } } },
            .{ .key = "service.version", .value = .{ .value = .{ .string_value = self.service.version } } },
            .{ .key = "service.namespace", .value = .{ .value = .{ .string_value = self.service.namespace } } },
        };
```

Is correct and matches the other proto syntax one may see in something like [golang](https://github.com/open-telemetry/opamp-go/blob/edef68087b2e443511cf37ccdee0f3950c5e6ddf/internal/examples/agent/agent/agent.go#L239-L243). I did find, however, that when encoding to JSON, the extra value field is included which causes an error when reading a proto message in other languages:

```json
[
  {
    "key": "service.name",
    "value": {
      "value": {
        "stringValue": "example"
      }
    }
  },
  {
    "key": "service.instance.id",
    "value": {
      "value": {
        "stringValue": "example-1764783202368-14467194"
      }
    }
  },
  {
    "key": "service.version",
    "value": {
      "value": {
        "stringValue": "latest"
      }
    }
  },
  {
    "key": "service.namespace",
    "value": {
      "value": {
        "stringValue": "example"
      }
    }
  }
]
```

as opposed to the expected like [this JSON example](https://github.com/open-telemetry/opentelemetry-proto/blob/a5f0eac5b802f7ae51dfe41e5116fe5548955e64/examples/trace.json#L5-L12) elsewhere in otel. 

I am encoding like so: 
```zig
const request_body = try request.jsonEncode(.{}, self.allocator);
        defer self.allocator.free(request_body);
```

Let me know if there's something I'm doing incorrect here, but I believe this is a bug in the json serialization.

inge4pres:
### Reason for this PR

Closes #147 
Supersedes: #148 

We want to give users ability to discard the nested container value in JSON encoding/decoding.

Note for reviewers: don't get scared by the large diff, **reviewing by commit should be doable**. The second commit has an update to the exiting API that cascades to all generated files.
Let me know if you prefer to keep the second commit or if we should split the API adding a "WithOptions" public variant.

### Details

Split by commit:

1. Add an `Options` struct to configure the encoding/decoding behavior
2. Change the `json.stringify` API to add an optional parameter passing user-provided Options (mouthful not intended 😅)

### DIFF
--- a/tests/tests_json.zig
+++ b/tests/tests_json.zig
@@ -1023,3 +1023,136 @@ test "JSON: decode selfref structs" {
 test "JSON: encode selfref structs" {
     // TODO
 }
+
+// OneofContainer flat-format (issue #147) tests
+const oneof_zig = @import("./generated/tests/oneof.pb.zig");
+const OneofContainer = oneof_zig.OneofContainer;
+
+test "JSON: encode oneof flat format (emit_oneof_field_name=false)" {
+    const msg = OneofContainer{
+        .regular_field = "hello",
+        .some_oneof = .{ .string_in_oneof = "world" },
+    };
+    // Use a wrapper so std.json.Stringify.valueAlloc calls our stringifyTo
+    const Wrapper = struct {
+        ptr: *const OneofContainer,
+        pub fn jsonStringify(self: *const @This(), jws: anytype) !void {
+            return protobuf.json.stringifyWithOptions(
+                OneofContainer,
+                self.ptr,
+                jws,
+                .{ .emit_oneof_field_name = false },
+            );
+        }
+    };
+    const encoded = try std.json.Stringify.valueAlloc(
+        allocator,
+        Wrapper{ .ptr = &msg },
+        .{},
+    );
+    defer allocator.free(encoded);
+    // Flat format: no "someOneof" wrapper key; variant key directly in parent object
+    try expect(std.mem.indexOf(u8, encoded, "someOneof") == null);
+    try expect(std.mem.indexOf(u8, encoded, "stringInOneof") != null);
+    try expect(std.mem.indexOf(u8, encoded, "regularField") != null);
+}
+
+test "JSON: encode oneof legacy wrapped format (emit_oneof_field_name=true)" {
+    var pb_instance = try string_in_oneof_init(allocator);
+    defer pb_instance.deinit(allocator);
+    // Default jsonEncode uses wrapped format (backward compat)
+    const encoded = try pb_instance.jsonEncode(.{}, allocator);
+    defer allocator.free(encoded);
+    // Legacy format: "someOneof" wrapper key present
+    try expect(std.mem.indexOf(u8, encoded, "someOneof") != null);
+    try expect(std.mem.indexOf(u8, encoded, "stringInOneof") != null);
+}
+
+test "JSON: decode oneof flat format (variant key directly in parent)" {
+    // Flat format: variant name at top level, no "someOneof" wrapper
+    const json_str =
+        \\{"regularField":"hello","stringInOneof":"world"}
+    ;
+    const result = try std.json.parseFromSlice(
+        OneofContainer,
+        allocator,
+        json_str,
+        .{},
+    );
+    defer result.deinit();
+    try std.testing.expectEqualStrings("hello", result.value.regular_field);
+    try expect(result.value.some_oneof != null);
+    try std.testing.expectEqualStrings("world", result.value.some_oneof.?.string_in_oneof);
+}
+
+test "JSON: decode oneof flat format with camelCase variant name" {
+    const json_str =
+        \\{"stringInOneof":"camelTest"}
+    ;
+    const result = try std.json.parseFromSlice(
+        OneofContainer,
+        allocator,
+        json_str,
+        .{},
+    );
+    defer result.deinit();
+    try expect(result.value.some_oneof != null);
+    try std.testing.expectEqualStrings("camelTest", result.value.some_oneof.?.string_in_oneof);
+}
+
+test "JSON: decode oneof flat format with snake_case variant name" {
+    const json_str =
+        \\{"string_in_oneof":"snake_test"}
+    ;
+    const result = try std.json.parseFromSlice(
+        OneofContainer,
+        allocator,
+        json_str,
+        .{},
+    );
+    defer result.deinit();
+    try expect(result.value.some_oneof != null);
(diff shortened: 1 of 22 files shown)


Arwalk (commented):
Woudln't it be possible to have the options exposed in the protobuf api?

i think that we could have something like:

generated file: 
```zig
pub fn jsonEncode(
        self: @This(),
        stdJsonOptions: std.json.Stringify.Options,
        protobufJsonOptions : protobuf.json.Options,
        allocator: std.mem.Allocator,
    ) ![]const u8 {
        return protobuf.json.encode(self, stdJsonOptions, protobufJsonOptions, allocator);
    }
```

json.zig:
```zig
pub fn encode(
    data: anytype,
    stdJsonOptions: std.json.Stringify.Options,
    protobufJsonOptions : protobuf.json.Options,
    allocator: std.mem.Allocator,
) ![]u8 {
    const Wrapper = struct {
        data : @TypeOf(data),
        pbJsonOptions : protobuf.json.Options,

         pub fn jsonStringify (self: *const @This(), jws: anytype) !void {
            return protobuf.json.stringify(@This(), self, jws, self.pbJsonOptions);
         }
    }
    
    return std.json.Stringify.valueAlloc(allocator, Wrapper{.data = data, .pbJsonOptions = pbJsonOptions},  stdJsonOptions);
}

pub fn stringify(Self: type, self: *const Self, jws: anytype, opts: ?Options) !void {
   const data = self.data;
   const  pbJsonOptions = self.pbJsonOptions;
   // work with that 
}
```

This might even allow us to remove the jsonStringify from the generated structs. 
I'm not sure if this works, this is just a theory. What do you think @inge4pres  ?

inge4pres:
@Arwalk that seems reasonable, and I think it's a great idea! Because right now using `encode` is the right way to get a JSON, and `stringify` is a specialization of that.
It comes with a change in public-facing API but I think that's tolerable as long as the minor version is increased.

I noticed there is additional logic that there is in `stringify` compared to `encode`. I think we should keep that logic (which does camel_case conversion and with this PR oneOf removal) and move it inside `encode` if we want to remove the duplication, but that comes with a breaking change correct?

Arwalk:
I'm ok with breaking changes in the JSON interfaces. As stated in the README, it's still considered beta. Still, we'll increase the major version accordingly (as it will be a non-retro-compatible interface change and we're doing our best to stay true to [semantic versioning](https://semver.org/))

Thanks again @inge4pres !

inge4pres:
> I'm ok with breaking changes in the JSON interfaces.

@Arwalk nice thanks, LGTM I will proceed if you can confirm my assumption in the previous comment

> I think we should keep that logic (which does camel_case conversion and with this PR oneOf removal) and move it inside encode if we want to remove the duplication, but that comes with a breaking change correct?

Arwalk:
> I think we should keep that logic (which does camel_case conversion and with this PR oneOf removal) and move it inside encode if we want to remove the duplication, but that comes with a breaking change correct?

Sorry i wasn't clear on that. Yes, you can go ahead. Thanks again for everything.

inge4pres:
It's done! 🚀 
I split in 2 more commits the changes described above.

**This is a breaking change** so mind it when merging, a new major version upgrade will be needed.

tests/tests_json.zig
```diff
@@ -1023,3 +1038,105 @@ test "JSON: decode selfref structs" {
...
+    const msg = OneofContainer{
+        .regular_field = "hello",
+        .some_oneof = .{ .string_in_oneof = "world" },
+    };
+    const encoded = try msg.jsonEncode(.{}, .{ .emit_oneof_field_name = false }, allocator);
+    defer allocator.free(encoded);
+    // Flat format: no "someOneof" wrapper key; variant key directly in parent object
+    try expect(std.mem.indexOf(u8, encoded, "someOneof") == null);
+    try expect(std.mem.indexOf(u8, encoded, "stringInOneof") != null);
+    try expect(std.mem.indexOf(u8, encoded, "regularField") != null);
```
Arwalk: Can we compare with entire json data instead of looking into strings? I like my tests to break when something changes and a big ol' expectEqualStrings is pretty good for checking if we're doing something wrong.

tests/tests_json.zig
```diff
@@ -1023,3 +1038,105 @@ test "JSON: decode selfref structs" {
...
+
+test "JSON: encode oneof legacy wrapped format (emit_oneof_field_name=true)" {
+    var pb_instance = try string_in_oneof_init(allocator);
+    defer pb_instance.deinit(allocator);
+    // Default jsonEncode uses wrapped format (backward compat)
+    const encoded = try pb_instance.jsonEncode(.{}, .{}, allocator);
+    defer allocator.free(encoded);
+    // Legacy format: "someOneof" wrapper key present
+    try expect(std.mem.indexOf(u8, encoded, "someOneof") != null);
+    try expect(std.mem.indexOf(u8, encoded, "stringInOneof") != null);
```
Arwalk: same as previous comment

tests/tests_json.zig
```diff
@@ -1023,3 +1038,105 @@ test "JSON: decode selfref structs" {
...
+test "JSON: roundtrip oneof flat format" {
+    const original = OneofContainer{
+        .regular_field = "roundtrip",
+        .some_oneof = .{ .string_in_oneof = "value" },
+    };
+    const encoded = try original.jsonEncode(.{}, .{ .emit_oneof_field_name = false }, allocator);
+    defer allocator.free(encoded);
+
+    // Encoded JSON should not contain "someOneof" wrapper
+    try expect(std.mem.indexOf(u8, encoded, "someOneof") == null);
```
Arwalk: same comment as before

Arwalk (changes requested):
Minor changes on the tests but overall this is a great job.

inge4pres:
> Minor changes

Feedback incorporated 🚀

Arwalk (approved):

Arwalk:
Really cool work. I really like how passing custom options is handled here. I hope i'll find the time to make a blog post or something for other people that'd like to have custom informations in their json encoding.

Thanks @inge4pres , i'll do a release soonish

PR #165, opened 2026-03-29
Include the package name in generated service struct

agagniere:
Hi,

In order to generate the client-side stub from the generated VTable at comptime, we need the package name, to build the fully qualified name `/package.service/Method`.

### DIFF
--- a/bootstrapped-generator/main.zig
+++ b/bootstrapped-generator/main.zig
@@ -990,6 +990,14 @@ const GenerationContext = struct {
                 allocator,
                 try std.fmt.allocPrint(
                     allocator,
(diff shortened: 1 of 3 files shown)


menduz (approved):
Please share more details about what you are cooking!

agagniere:
I'm using this for my grpc wrapper: [example that uses the package name](https://github.com/agagniere/cgrpc_wrapper/blob/master/examples/greeter_stub.zig) to hide away protobuf's encode/decode

For now only unary client calls are supported because that's the minimum needed for sending OpenTelemetry data.
I will tackle streams next

PR #171, opened 2026-05-19
Conformance tests fixes (Part 2 of 2)

menduz:
This PR fixes all conformance tests. Most of the tests are around google's well-known types. i.e. Timestamp. Those have very specific json parsing and emitting, while preserving an underlying Message with native types. Those mechanisms allow for things like parsing a ISO timestamp or a i64 in the shape of a string encoded number (because JSON only allows for doubles, and otherwise we may lose resolution).

Even though those datatypes are discouraged, we want the library to support them. The parsing of those datatypes lives in a brand new file called `wkt.zig`, and the generator references them when needed. *Those parsers are not battle tested, and yours truly (and google) discourages the usage of this built-in datatypes.* Like the rest of this library, those work without any external library. But if you are worried about increasing the surface area of attack of your application, the changes are contained only to the json (de)serializers.

To make the conformance tests pass, I've also fixed the recursive and corecursive messages scenarios.

I'd like to get the review from the folks doing opentelemetry before merging this. The results adhere to what protocolbuffers (the lib) says the JSON results should be, but that differs from the golden file tests that we've got in place (see the diff). 

@agagniere @inge4pres 

```
CONFORMANCE TEST BEGIN ====================================

CONFORMANCE SUITE PASSED: 1476 successes, 4091 skipped, 0 expected failures, 0 unexpected failures.

CONFORMANCE TEST BEGIN ====================================

CONFORMANCE SUITE PASSED: 0 successes, 883 skipped, 0 expected failures, 0 unexpected failures.

info: conformance-zig-protobuf: served 883 tests
info: conformance-zig-protobuf: served 5567 tests
```


bootstrapped-generator/main.zig
```diff
@@ -100,16 +101,50 @@ const GenerationContext = struct {
...
 
     pub fn init(allocator: std.mem.Allocator, request: plugin.CodeGeneratorRequest) !GenerationContext {
-        return .{
+        var ctx: GenerationContext = .{
             .req = request,
             .res = .{},
             .known_packages = .init(allocator),
             .fqn_lines = .init(allocator),
             .message_deps = .init(allocator),
+            .preserve_unknown_fields = false,
```
menduz, line 110: This PR introduces an optional `[]const u8` accumulator for unknown fields

bootstrapped-generator/main.zig
```diff
@@ -264,6 +299,51 @@ const GenerationContext = struct {
...
+        "google.protobuf.UInt64Value",
+        "google.protobuf.Int32Value",
+        "google.protobuf.UInt32Value",
+        "google.protobuf.BoolValue",
+        "google.protobuf.StringValue",
+        "google.protobuf.BytesValue",
+    };
+    const wkt_enum_fqns = [_][]const u8{
+        "google.protobuf.NullValue",
+    };
```
menduz, line 324: Implementation for these Messages contains custom serializers, we point to wkt.zig's Any instead of generating google.protobuf.Any. In a future we may want to enable custom hand-made serializers as parameter for the compiler.

bootstrapped-generator/main.zig
```diff
@@ -1196,6 +1353,7 @@ fn isScalarNumeric(t: descriptor.FieldDescriptorProto.Type) bool {
         .TYPE_SFIXED32,
         .TYPE_SFIXED64,
         .TYPE_BOOL,
+        .TYPE_ENUM,
```
menduz, line 1356: This was a real bug!

src/wire.zig
```diff
@@ -1087,14 +1153,100 @@ fn skipField(reader: *std.Io.Reader, tag: Tag) !usize {
             _, const c = try decodeScalar(.uint64, reader);
             consumed += c;
         },
-        .sgroup, .egroup => {
-            // `sgroup`/`egroup` not supported.
+        .sgroup => {
```
menduz, line 1156: The parser now supports (and correctly ignore) groups

tests/generated/jspb/test.pb.zig
```diff
@@ -2244,7 +2244,7 @@ pub const TestMapFieldsNoBinary = struct {
 
     pub const MapStringTestmapfieldsEntry = struct {
         key: ?[]const u8 = null,
-        value: ?TestMapFieldsNoBinary = null,
+        value: ?*TestMapFieldsNoBinary = null,
```
menduz, line 2247: This one may be a bug. Open to discuss. It is not strictly necessary to use the pointer here. But if we are purists, it makes sense, as the entry of the map (represented as an array) is recursive.

```protobuf
message TestMapFieldsNoBinary {
  map<string, TestMapFieldsNoBinary> map_string_testmapfields = 12;
  // roughly translates to list<MapStringTestmapfieldsEntry> map_string_testmapfields = 12;
}
```

tests/json_data/oneof_container/message_in_oneof_camelCase.json
```diff
@@ -1,6 +1,5 @@
 {
   "regularField": "this field is always the same",
-  "enumField": "UNSPECIFIED",
```
menduz, line 3: proto3 removes default values from json serializations

menduz:
@agagniere @inge4pres I'm planning to merge this early next week

agagniere (approved):
I don't know protobuf deeply enough to have an opinion on the well-known-types JSON serialization so I reviewed what I could.

Only nit is the date conversions using magic numbers, proposed patch: https://github.com/agagniere/zig-protobuf/commit/671ae302835e7f45ae1d6470f58259bf24bf852f

Arwalk:
Thank you guys for all of this, i'll review this asap

Arwalk (approved):
Note: On the json map encoding tests: I prefer when we check the output with an actual JSON string, and not something that goes through another parser. But that's not blocking.

Dis gud !

menduz:
Will add unit tests for the conformance of editions 2023

PR #173, opened 2026-05-25
fix: support platforms without prebuilt protoc binaries

tuaris:
On platforms where no prebuilt protoc binary is available (e.g. FreeBSD, OpenBSD, NetBSD), getProtocDependency() previously called @panic(). This made it impossible to use zig-protobuf as a dependency on any of these systems.

Changes:
- getProtocDependency(): return null instead of @panic on unsupported platforms, allowing the caller to handle the missing dependency gracefully.
- getProtocBin(): when no prebuilt binary is available, search for a system-installed protoc in /usr/local/bin and /usr/bin before giving up.

This enables zig-protobuf (and all downstream packages such as opentelemetry-sdk) to build on FreeBSD and other BSDs where protoc is installed via the system package manager (e.g. devel/protobuf on FreeBSD).

Tested on FreeBSD 15.0 with Zig 0.15.2 and system protoc 29.5.

### DIFF
--- a/build_util.zig
+++ b/build_util.zig
@@ -36,6 +36,7 @@ pub fn getProtocDependency(b: *std.Build) !?*std.Build.Dependency {
     const os: ?[]const u8 = switch (builtin.os.tag) {
         .macos => "osx",
         .linux => "linux",
+        .freebsd, .openbsd, .netbsd, .dragonfly => "bsd",
         else => null,
     };
 
@@ -50,6 +51,8 @@ pub fn getProtocDependency(b: *std.Build) !?*std.Build.Dependency {
 
     const dependencyName = if (builtin.os.tag == .windows)
         try std.mem.concat(b.allocator, u8, &.{"protoc-win64"})
+    else if (os != null and std.mem.eql(u8, os.?, "bsd"))
+        return null // BSDs use system-installed protoc
     else if (os != null and arch != null)
         try std.mem.concat(b.allocator, u8, &.{ "protoc-", os.?, "-", arch.? })
     else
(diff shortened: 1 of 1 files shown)


tuaris:
How can we backport this to Zig 0.15.x, which is your v4 tag?

menduz:
I think you should use `RunProtoc.compile(.{ .protoc_override = ... })` for this use case

Arwalk:
> How can we backport this to Zig 0.15.x, which is your v4 tag?

The best I can do would be to create a zig-0.15.2 branch, but i won't be able to make any maintenance on it. It is honestly easier for you to fork on this subject.

For current zig release, there is the option provided by @menduz that you could consider and try. I'm also working on making protoc potentially built on the spot for other platforms.

build_util.zig
```diff
@@ -70,6 +73,24 @@ pub fn getProtocBin(protoc_owner: *std.Build, step: *std.Build.Step) !?[]const u
...
+    // No prebuilt protoc binary for this platform.
+    // Fall back to a system-installed protoc.
+    return findSystemProtoc(step.owner.graph.io);
+}
+
+/// Search for a system-installed protoc binary in common locations.
+/// This enables building on platforms without prebuilt protoc binaries
+/// (e.g. FreeBSD, OpenBSD, NetBSD) as long as protoc is installed via
+/// the system package manager.
+fn findSystemProtoc(io: Io) ?[]const u8 {
```
Arwalk, line 86: Wouldn't it be better to run `which protoc` on these platforms ?

menduz:
@tuaris please check https://github.com/Arwalk/zig-protobuf/commit/00dfee67c1c0db5e198d82e69872c3d567bfb000

PR #174, opened 2026-05-28
Improve JSON serialization support

Jeff-Ciesielski:
I apologize for the size of this PR, I set out to fix the JSON map field encoding as I need it imminently for a project and got a little carried away.  I've tried to split this into functional commits that can stand alone. But there all fairly related on theme so I figured killing it all in one shot might be acceptable.

If you'd prefer that I break this down into a few individual PRs, I'd be happy to do so, just say the word, I don't want to create undue burden.


menduz:
Hello Jeff, many thanks for putting this together. I think we both addressed the same things. Here is my PR, additionally it fixes some bugs and makes the conformance tests pass.  https://github.com/Arwalk/zig-protobuf/pull/171

Jeff-Ciesielski:
@menduz Hah! Oh damn, well I'm glad it'll get in either way!  I'll close this in favor of yours

menduz:
I've adopted your unit tests, thanks! @Jeff-Ciesielski

Arwalk:
Thank you for your contributions @Jeff-Ciesielski , even if it wasnt merged.

PR #175, opened 2026-05-31
fix(dupe): use allocator.dupe for repeated scalar string/bytes fields

hanapedia:
Items in a `repeated` field with scalar types of `.string` or `.bytes` have no `.dupe` method. This would fail to compile for any message containing repeated sting or repeated bytes field if `.dupe()` was called. This commit fixes this bug by using `allocator.dupe` instead.

I've also added tests that would have failed to compile just as described above if the fix were not applied.


menduz (approved):
Thank you!

Arwalk:
Thank you!

PR #176, opened 2026-06-04
feat: wire performance improvements

jaronoff97:


src/json.zig
```diff
@@ -11,8 +11,29 @@ pub const Options = struct {
...
+/// Thread-local toggle for hex-encoding `bytes` fields. std.json's parse/stringify
+/// machinery can't thread custom options through its recursion, so this mirrors the
+/// thread-local used for the `Any` allocator (`wkt.tl_any_alloc`).
+///
+/// `encode` sets this automatically from `Options.bytes_as_hex`. Decode callers that
+/// need hex bytes must set it directly before invoking `decode`/`jsonDecode`:
+///
+///   protobuf.json.tl_bytes_as_hex = true;
+///   defer protobuf.json.tl_bytes_as_hex = false;
+pub threadlocal var tl_bytes_as_hex: bool = false;
```
menduz: style: while you are still at it, please do not include thread locals or globals. prefer passing arguments

menduz: oh, wkt.any already does. I've created [an issue](https://github.com/Arwalk/zig-protobuf/issues/177)

jaronoff97: yeah sorry, i want to get rid of this entirely, I do this because Otel protos have some funky stuff around hex as bytes. I'm working through in this very moment if i can get rid of threadlocal and this option as I really dislike it haha

menduz: if you can, changing the datatype of the field from `bytes` to protobufs WKT bytes `google.protobuf.BytesValue`, the primitive will still be bytes but its serialization will be hex, for free. you would only need to hand tune the .proto and re-emit the code

jaronoff97: unfortunately this isn't so simple as this is the upstream proto these are generated from https://github.com/open-telemetry/opentelemetry-proto/blob/82dd303497e1bfbc86f1f8d584696195a02ea085/opentelemetry/proto/trace/v1/trace.proto#L86-L104

src/wire.zig
```diff
@@ -469,10 +487,15 @@ pub fn decodeRepeated(
...
-            try result.append(
-                allocator,
-                try protobuf.init(Result, allocator),
-            );
+            // Ensure capacity before append to reduce reallocations. When at
+            // capacity, grow by at least 8 (or double) to amortize allocation
+            // cost across repeated submessage occurrences.
+            if (result.items.len >= result.capacity) {
+                const grow_by = @max(8, result.capacity);
+                try result.ensureTotalCapacity(allocator, result.capacity + grow_by);
```
agagniere, line 495: Nit:
```suggestion
                try result.ensureUnusedCapacity(allocator, grow_by);
```
(in this `if`, `result.items.len == result.capacity`)

src/wire.zig
```diff
@@ -469,10 +487,15 @@ pub fn decodeRepeated(
             // Submessages are length-delimited, and cannot be packed.
             std.debug.assert(options.bytes != null);
 
-            try result.append(
-                allocator,
-                try protobuf.init(Result, allocator),
-            );
+            // Ensure capacity before append to reduce reallocations. When at
+            // capacity, grow by at least 8 (or double) to amortize allocation
+            // cost across repeated submessage occurrences.
```
agagniere, line 492: (ArrayList is already amortizing allocation cost, but indeed I see you are forcing a growth factor of >= 2.
Surprisingly, when `grow_by = result.capacity`, [`ensureTotalCapacity`](https://ziglang.org/documentation/master/std/#std.array_list.Aligned.ensureTotalCapacity)'s `new_capacity` is `2 * capacity`,  and so [`growCapacity`](https://ziglang.org/documentation/master/std/#std.array_list.Aligned.growCapacity) will return `3 * capacity + 128 / @sizeOf(T)`, so the growth factor will actually be >= 3)

agagniere (approved):

Issue #30, opened 2023-12-31
Incremental decoding / Streaming API


Rubeer:
While the protobuf encoding allows you to decode one field at a time at a low level, it is usually more convenient to have an API where you decode one message at a time.

To do this, most protobuf libraries will require the entire encoded message to be in contiguous memory before decoding it. This is fine in most cases, but sometimes you want to be able to decode messages incrementally. This saves on copying out to intermediate buffers.

Example use case in an embedded system, or a kernel driver:
Data is received into a ringbuffer or FIFO from some hardware peripheral. This buffer could be smaller than the size of an encoded protobuf message, and depending on the head/tail position, some messages may also wrap around from the end to the start of the buffer. Normally it would need to be defragmented/copied to an intermediate buffer before decoding, which takes more memory, or you need to resort to manually parsing one field at a time. In any case some manual logic / state machine is required.

Proposal: a "throw bytes at this function" API, where the library calls you back every time it has fully decoded a message.
The decoding state machine is generated by the library.
Example usage code:
```rs
// Only need to call this once. Possibly comptime.
var decode_state = protobuf.pb_decoder_init(.{ 
    .message_type = MyMessageType,
    .allocator = allocator,
    .delimiter_parser = protobuf.VarIntDelimiterParser, // Library provided implementation, or could be a custom implementation
    .decode_callback = message_decoded_callback,
});

// Somewhere else in the codebase: we receive some bytes.
// This is a slice with any number of bytes, i.e. half a message, or three entire messages.
// message_decoded_callback is called whenever a MyMessageType is fully decoded,
// across any number of calls to pb_decode_bytes.
// This function should only use a buffer large enough to decode one field.
// It only calls the allocator when decoding dynamic data like strings or repeated fields.
try protobuf.pb_decode_bytes(&decode_state, bytes_from_somewhere());


fn message_decoded_callback(decoded: *const MyMessageType) void {
    defer decoded.deinit();

    // do stuff with decoded message

    
}
```

I am not familiar enough with zig-protobuf internals to know how hard this to implement, but I imagine it would be a good chunk of code. From an API design perspective the hard part is going to be the delimiter parser, which would be some interface to parse the delimiters/headers. Ideally it would be flexible enough that you can implement things like checksums/CRC with it as well.

A streaming encoder would be nice to have as well, but it is not as important in my opinion.

I ended up typing out quite the story here, but it is just some draft/idea, I understand if it is outside of the scope of this project.

[...]

PR #179, opened 2026-06-08
feat(stream): added stream decoder

Arwalk:


src/stream.zig
```diff
@@ -0,0 +1,611 @@
+//! Streaming pull-decoder for protobuf messages.
+//!
+//! Where `protobuf.decode` materializes a whole message (allocating storage for
+//! every dynamic field), `StreamDecoder` walks a `std.Io.Reader` one wire field
+//! at a time and hands each one back as an `Event`. It never allocates: scalar
+//! values are returned by value, and length-delimited fields (submessages,
+//! strings, bytes) are surfaced as a `*std.Io.Reader` bounded to that field's
```
menduz, line 7: is the ram returned by streams safe to use after the reader is consumed?

Who owns that slice? the socket? the filesystem? It may be required that after producing a message, the entire message needs to get duped using a user-provided allocator..

Arwalk: This is still a draft for a reason :)

README.md
```diff
@@ -83,6 +83,43 @@ zig-protobuf generates code for Protocol Buffer `service` definitions using the
...
+Besides `MyMessage.decode`, which materializes a whole message (allocating storage for
+every dynamic field), every generated message also exposes a `StreamDecoder`: a
+zero-allocation **pull parser** that walks a `std.Io.Reader` one wire field at a time.
+This is useful for incremental / low-memory decoding — large or deeply nested messages,
+embedded systems, or multiplexed IO — where you don't want to buffer a whole message in
+contiguous memory.
+
+Call `next()` to get the next field as an `Event`. Scalars come back by value; the leaf
+cases of a `oneof` are flattened into their own variants. Length-delimited fields
+(submessages, `string`, `bytes`) are surfaced as a `*std.Io.Reader` bounded to that
```
menduz:
```suggestion
(submessages, `string`, `bytes`) are surfaced as a `*std.Io.Reader` bound to that
```

Arwalk:
note: important footgun that must appear in the documentation is that this may mean that a decoding error in the middle of the decoding stream may implies that some processing has been partially applied

src/stream.zig
Arwalk: i've added more comments overall. Is that enough?

src/stream.zig
```diff
@@ -0,0 +1,677 @@
...
+//! arbitrarily large or nested messages without buffering them in contiguous
+//! memory and without the library taking over control flow.
+//!
+//! Usage:
+//! ```zig
+//! var sd = MyMessage.StreamDecoder.init(&reader);
+//! while (try sd.next()) |item| switch (item) {
+//!     .some_scalar => |v| { ... },
+//!     .some_submessage => |limited| {        // limited: *std.Io.Reader
+//!         var inner = Sub.StreamDecoder.init(limited);
```
menduz, line 20: I wonder what happens if the limited stream is not yet complete

menduz:
I think I'll need to know more context and motivations to correctly review this PR

Arwalk:
Sure, no problem. First, there is this issue here #30 which gives a concrete case:

> Example use case in an embedded system, or a kernel driver:
> Data is received into a ringbuffer or FIFO from some hardware peripheral. This buffer could be smaller than the size of > an encoded protobuf message, and depending on the head/tail position, some messages may also wrap around from > the end to the start of the buffer. Normally it would need to be defragmented/copied to an intermediate buffer before > decoding, which takes more memory, or you need to resort to manually parsing one field at a time. In any case some > manual logic / state machine is required.

Apart from this, as someone who mostly work in memory-constrained environment with no dynamic memory allocation _allowed_, the need to have an interface that doesn't take any decisions on how to keep the data around is useful. I use [nanopb](https://github.com/nanopb/nanopb) in these environments because its interface allows to work that way, using callbacks when repeated/len-delimited fields are met. (see https://jpa.kapsi.fi/nanopb/docs/concepts.html#decoding-callbacks )

I didn't want to make a callback-based thing on zig-protobuf as we can have much better (and cooler 😎 ) interface with comptime.

For small embedded environments that must not waste resources, an API like this allows to process data while it is being decoded, without waiting to decode the entire chain.

Here a real use case : i have a device that receives [APDU](https://en.wikipedia.org/wiki/Smart_card_application_protocol_data_unit) to send to a smartcard, as well as the expected status word (think return code). Using the streaming api of nanopb, the device sends them by progressively decoding each APDU request. If any fail, it can stop right away in its process and returns quickly to the master device that something went wrong.

I know it's not the most common use case, but protobuf being small makes it perfect for small systems with performance requirements means that it is actually used in these context. That's personally how I use it at work.

src/stream.zig
Arwalk: You mean not used entirely? Or it can't get any more byte because of lag/delay/whatever ?

Arwalk:
[gremlin.zig](https://github.com/norma-core/gremlin.zig) also has a stream with no allocator decoder. The need exists!

menduz:
Haven't forgot about this

menduz (approved):
Changeset looks reasonable, but until I see some production code, I think that manually iterating through the read fields recursively would be better that iterating through fields at shallow level of the first decoder.

I can use this in a project in a couple weeks and try it, but I doubt it will make sense for me because this project has deeply nested messages. But I think that the stream decoder makes sense for embedded applications. 

My opinion on the code is a _go_, my opinion on the ergonomics is that fields at top level message are the only ones confortable enough to be usable. This is only possible because protobuf not-packed repeated fields are simply re-emited with their tag, and we can process a stream from a server flawlessly and emit fields and repeated fields with an union.

```protobuf
message FromServerUnion {
  oneof my_union {
    AuthenticationResponse auth = 1;
    ChatNotification chat = 2;
    KickedNotification kicked = 3;
    UserJoinedNotification user+joined = 4;
    UserLeftNotification user_left = 5;
  }
}
message FromServerStream {
  repeated FromServerUnion msg_union = 1;
}
```

```zig

test "stream: submessage unions wait for the whole submessage to be completed before emitting" {
    // emits half a message to emulate a split TCP packet
    var writer = InMemoryChunkedMultiThreadWriter()
    var reader = ChunkedReaderWithDelay(writer);

    var sd = StreamDecoder(FromServerStream).init(&reader);

    {
      writer.writeInPartsWithDelay(FromServerStream{ .stream = &.{ .auth = {} } });
      const serverMessage: FromServerUnion = (try sd.next()).?.msg_union;
      // assert serverMessage.auth is a complete message
    }

;

    {
      writer.writeInPartsWithDelay(FromServerStream{ .stream = &.{ .auth = {} } });
      const serverMessage: FromServerUnion = (try sd.next()).?.msg_union;
      // assert serverMessage.auth is a complete message
    }
}

```

Arwalk:
>  But I think that the stream decoder makes sense for embedded applications.
> My opinion on the code is a go, my opinion on the ergonomics is that fields at top level message are the only ones confortable enough to be usable

In my own experience, embedded application do not have complicated messages. I agree with you that it wouldn't make a lot of sense to use the stream decoder for complicated messages: if you can already afford this complexity you're usually out of the realm of the embedded world.

PR #180, opened 2026-06-15
Simplify public facing compiler API, enable custom protoc executable

menduz:
Simplifies and fixes the public facing API for build.zig, we were using a wrong owner and that broke the README example.

Hardly used fields were removed in favor of passing `LazyPath("protoc")` which I consider more useful for people that 1) do not want to use google's 2) there are no prebuilt binaries (BSD)

I think this PR can also be enough to close https://github.com/Arwalk/zig-protobuf/pull/173


Arwalk (approved):
Dis gud.

Side note: i had https://github.com/zig-protobuf/zig-protoc working at some point, i should get back into it to propose compiled on the spot protoc.

menduz:
We do compile it on the spot in the conformance tests if you want to revamp it

PR #181, opened 2026-06-20
build(evalBranchQuota): dynamically calculate a minimum value for evalBranchQuota

### LINKED ISSUE #152: Handle eval branches quota more gracefully
The whole point of this library is to rely as much as possible on comptime, but we have hit multiple times the eval branch quota.

Just adding `@setEvalBranchQuota(1000000);` is a quick and very dirty way to get out of it, but there should be a better way to do this, as i'm pretty sure that very complex protobuf files could hit the limit.

The only idea that i have so far would be to try to use some kind of build option to pass an eval limit to the rest of the code. (see the [documentation](https://ziglang.org/learn/build-system/#options-for-conditional-compilation) for a rough idea). That way the end user can fix the limit itself if his use case breaks.

Any other idea?

Arwalk:
@setEvalBranchQuota accepts any comptime value. That means it can be computed at comptime too. By (ab)using this, we can parse the protobuf structures at the entry point of all public functions and set a value for @setEvalBranchQuota depending on the number of fields that can be reached in a structure and its sub structures.

closes #152

### DIFF
--- a/src/json.zig
+++ b/src/json.zig
@@ -19,8 +19,7 @@ pub fn parse(
     source: anytype,
     options: std.json.ParseOptions,
 ) !Self {
-    // Increase eval branch quota for types with hundreds of fields
-    @setEvalBranchQuota(1000000);
+    @setEvalBranchQuota(comptime protobuf.evalBranchQuotaFor(Self));
 
     if (.object_begin != try source.next()) {
         return error.UnexpectedToken;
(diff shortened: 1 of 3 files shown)


menduz (approved):

menduz:
I think the gauss curve meme applies for this PR

Arwalk:
> I think the gauss curve meme applies for this PR

I won't ask on which side of the curve i am, just to be safe.

Arwalk:
Fine, here it is. 
image

Issue #183, opened 2026-06-22
Make available the option to compile protoc


Arwalk:
In the conformance test, we compile protoc ourselves. This option should be exposed for end users that would prefer to use it that way, and not a binary blob. (see #144)

PR #185, opened 2026-07-01
WIP: build: use protobuf port instead of downloading binaries

Jan200101:
follow-up to #146

[allyourcodebase/protobuf](https://github.com/allyourcodebase/protobuf) now exists and builds protoc from source, this PR makes use of it for bootstrap but nothing else.

There are a lot of decisions to be made if this path is chosen such as:
- How will the dependency be shared around?
- How will the default include path be added (you already have dupeLazyPaths and might want to write another helper for it)
- etc.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -177,7 +177,8 @@ pub fn build(b: *std.Build) !void {
         test_step.dependOn(&run_main_tests.step);
     }
 
-    const include = if (try build_util.getProtocDependency(b)) |protoc| protoc.path("include") else b.path("");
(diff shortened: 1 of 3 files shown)


Arwalk:
hey there @Jan200101 , thank for all of this.

in #183, i note that we have already a place where we compile protoc locally, which are the conformance tests. I'm not sure we need an external dependency then. 

What i'd like though is exposing this built protoc to end users so they can work with it.

Do you think you'd like to do this?

In any case, thanks a lot for your contributions and continued interest.

